PR #279, opened 2023-05-30
doc: Added sequence diagrams

energinet-ajw:
## References

* #0000


docs/diagrams/sequence/Request change of supplier.txt
```diff
@@ -0,0 +1,23 @@
+autoactivation on
```
lasrinnil, line 1: Is there a generic implied way of displaying this txt file? :D

lasrinnil (approved):

PR #321, opened 2023-06-27
Fix: ensure that exception are being correctly logged in application …

MadsDue:
…insight

## References

* #0000


lasrinnil (approved):
Would love

PR #328, opened 2023-06-29
chore: EnableUserCodeException to see specific error type in applicat…

MadsDue:
EnableUserCodeException to see specific error type in application insigth instead of all exception as RpcException

https://github.com/Azure/azure-functions-dotnet-worker/pull/939


inishantmishra:
Even after enabling the EnableUserCodeException = true in dotnet isolated Azure function, I am still receiving RPC exceptions with no call stack.
I thought after implementing this change, if l throw some custom exception in my code and if I handle it and log it, I will get my custom exception in appInsights but it is not happening, I am still getting RPC exceptions with no call stack.


Let me know what am I missing? Why am i still receiving RPC exceptions?

I am using
dotnet-isolated
~4
.net 7 Isolated Azure function

PR #339, opened 2023-07-04
Removed check and test of message size exceeded.

energinet-ajw:
Since the API Management checks format (XML/JSON) and the size of the request, these validation checks have been removed from the solution.

## Description

https://app.zenhub.com/workspaces/phoenix--60a6105157304f00119be86e/issues/gh/energinet-datahub/team-phoenix/24

* #0000


lasrinnil (approved):
That's pretty damn great that we no longer have to check for size, great work!

PR #414, opened 2023-08-24
fix: updated process type and quality

RasmusGraabaek:
## Description

After an error was found, we have implemented a fix. In addition, we now handles 1) Logging when an integration event fails when processing it, 2) unittests that make sure that we map the correct properties from the contract CalculationResultCompleted in our CalculationMapper.

## References

* #0000


source/Domain/Transactions/Aggregations/SettlementVersion.cs
```diff
@@ -14,22 +14,19 @@
 
 using Domain.SeedWork;
 
-namespace Domain.OutgoingMessages;
+namespace Domain.Transactions.Aggregations;
 
-public class ProductType : EnumerationType
```
lasrinnil, line 19: Why are we overwriting `ProductType` with `SettlementVersion?`

MadsDue, line 19: Did we not use ProductType before?

source/Domain/Transactions/Aggregations/SettlementVersion.cs
```diff
@@ -14,22 +14,19 @@
...
+        Code = code;
     }
 
-    public static ProductType From(string valueToParse)
-    {
-        var productType = GetAll<ProductType>().FirstOrDefault(type => type.Name.Equals(valueToParse, StringComparison.OrdinalIgnoreCase)) ?? throw new InvalidOperationException($"{valueToParse} is not a valid product type");
-
-        return productType;
-    }
+    public string Code { get; }
```
lasrinnil, line 31: We are diverting from the pattern used for typed values following the `From` approach on the `EnumerationType` - why is this necessary?

source/Infrastructure/Transactions/Aggregations/CalculationResultCompletedEventMapper.cs
```diff
@@ -41,23 +41,28 @@ public async Task<INotification> MapFromAsync(string payload)
...
                 MapSettlementMethod(integrationEvent),
-                MapProcessType(integrationEvent),
+                MapProcessType(integrationEvent.ProcessType),
                 MapActorGrouping(integrationEvent),
-                await MapGridAreaDetailsAsync(integrationEvent).ConfigureAwait(false)));
+                await MapGridAreaDetailsAsync(integrationEvent).ConfigureAwait(false),
+                null,
+                null,
+                null,
+                MapSettlementVersion(integrationEvent)));
```
lasrinnil, line 59: Move this line to 56, then we don't have no parse `null` to the default paramteres

source/Infrastructure/Transactions/Aggregations/CalculationResultCompletedEventMapper.cs
```diff
@@ -66,70 +71,106 @@ public string ToJson(byte[] payload)
...
+            QuantityQuality.Measured => Quality.Measured.Name,
+            QuantityQuality.Missing => Quality.Missing.Name,
+            QuantityQuality.Estimated => Quality.Estimated.Name,
+            QuantityQuality.Calculated => Quality.Calculated.Name,
+            QuantityQuality.Unspecified => throw new InvalidOperationException("Quality is not specified"),
+            _ => throw new InvalidOperationException("Unknown quality type"),
+        };
+    }
+
+    private static string? MapSettlementMethod(CalculationResultCompleted integrationEvent)
```
lasrinnil, line 158: If you intend to change the signature of the supporting mapping method, I personally would prefer if it followed the same signature. So either use the enums or the event as a whole.

source/Tests/Domain/Transactions/Aggregations/CalculationResultCompletedEventMapperTests.cs
```diff
@@ -0,0 +1,143 @@
...
+    [Theory]
+    [MemberData(nameof(Resolutions))]
+    public void Ensure_handling_all_resolutions(Resolution resolution)
+    {
+        // Act
+        if (resolution != Resolution.Unspecified)
+        {
+            CalculationResultCompletedEventMapperSpy.MapResolutionSpy(resolution);
+        }
+        else
```
lasrinnil, line 89: The way I'm reading the implementation, this `else` case would ensure that even if new values were added to the enum it should guard, we wouldn't be made aware.

since all no-handled cases for some (I've not checked everyone) end with throwing an `InvalidOperationException` - then a new value would end up doing the same.

Have you tried the tests with a value wich was not handled and saw the test fail?

MadsDue (approved):

PR #420, opened 2023-08-25
fix: only storing transaction ids and message id, if the message is accepted

RasmusGraabaek:
## References

Problem: We get an internal server error, when we try to store Message ids and Transaction ids in the (Message/Transaction)Registry table, when they are over 50 characters.

It may be benificial to move all id saves to a db transaction. This approach may make it harder to give a propper error message, in cases where the transaction fails.


source/Api/IncomingMessages/RequestAggregatedMeasureMessageReceiver.cs
```diff
@@ -89,14 +105,38 @@ public class RequestAggregatedMeasureMessageReceiver
             return CreateResponse(request, httpErrorStatusCode, _responseFactory.From(errorResult, cimFormat));
         }
 
+        _messageArchive.Add(new ArchivedMessage(
+            Guid.NewGuid().ToString(),
```
lasrinnil, line 109: Not related to your work, but why is the Id of the ArchivedMessage set from the outside? Thats seems odd to me:)

source/Api/IncomingMessages/RequestAggregatedMeasureMessageReceiver.cs
```diff
@@ -89,14 +105,38 @@ public class RequestAggregatedMeasureMessageReceiver
             return CreateResponse(request, httpErrorStatusCode, _responseFactory.From(errorResult, cimFormat));
         }
 
+        _messageArchive.Add(new ArchivedMessage(
```
lasrinnil, line 108: Perhaps create a private method where you put the new message and the dbcontext save together?

source/Api/IncomingMessages/RequestAggregatedMeasureMessageReceiver.cs
```diff
@@ -89,14 +105,38 @@ public class RequestAggregatedMeasureMessageReceiver
...
+            Guid.NewGuid().ToString(),
+            messageHeader.MessageId,
+            IncomingDocumentType.RequestAggregatedMeasureData,
+            TryGetActorNumber(messageHeader.SenderId),
```
lasrinnil, line 112: Perhaps the concept of expecting a correct GLN or EIC number in this case is wrong? Maybe we just should accept the message header value id's?

source/Domain/ArchivedMessages/ArchivedMessage.cs
```diff
@@ -19,6 +19,9 @@
 namespace Domain.ArchivedMessages;
 
 public record ArchivedMessage(
+    // This is most likely a guid, but!
+    // We have examples of this being a 36 characters string not supporting the guid format in our DB.
+    // Hence we keep this as a string
```
lasrinnil, line 24: I would like to give a good explanation why I think we should remove this comment, tomorrow :)

source/Infrastructure/IncomingMessages/MessageIdRegistry.cs
```diff
@@ -30,17 +32,32 @@ public MessageIdRegistry(IDatabaseConnectionFactory connectionFactory)
...
+            {
+                throw new UnsuccessfulMessageIdStorageException();
+            }
+        }
+
+        public async Task<bool> MessageIdIsUniqueForSenderAsync(string senderId, string messageId, CancellationToken cancellationToken)
+        {
+            using var connection = await _connectionFactory.GetConnectionAndOpenAsync(cancellationToken).ConfigureAwait(false);
+            var message = await connection.QueryFirstOrDefaultAsync(
+                    $"SELECT * FROM dbo.MessageRegistry WHERE MessageId = @MessageId AND SenderId = @SenderId",
```
lasrinnil, line 56:
```suggestion
                    $"SELECT TOP 1 FROM dbo.MessageRegistry WHERE MessageId = @MessageId AND SenderId = @SenderId",
```

lasrinnil: Or something that makes sure that you only expect it to find one among the table

source/Infrastructure/IncomingMessages/MessageQueueDispatcher.cs
```diff
@@ -57,11 +57,7 @@ public Task AddAsync(IMarketTransaction message, CancellationToken cancellationT
 
         public async Task CommitAsync(CancellationToken cancellationToken)
         {
```
lasrinnil, line 59: Perhaps you could explain to me why this is correct tomorrow (Im pretty sure you are right, I just dont get it :)

source/Infrastructure/IncomingMessages/MessageQueueDispatcher.cs
```diff
@@ -57,11 +57,7 @@ public Task AddAsync(IMarketTransaction message, CancellationToken cancellationT
...
-            using (var scope = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled))
-            {
-                await _senderCreator.SendAsync(_transactionQueue.AsReadOnly(), cancellationToken).ConfigureAwait(false);
-                scope.Complete();
-            }
+            await _senderCreator.SendAsync(_transactionQueue.AsReadOnly(), cancellationToken).ConfigureAwait(false);
```
lasrinnil, line 60: Not related to your work!

What about if the _transactionQueue is to large to fit into a batch of ServiceBusMessages?

https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebussender.sendmessageasync?view=azure-dotnet

Think we need to solve this in another PR :)

source/Infrastructure/IncomingMessages/TransactionIdRegistry.cs
```diff
@@ -30,18 +34,56 @@ public TransactionIdRegistry(IDatabaseConnectionFactory connectionFactory)
...
                     new { TransactionId = transactionId, SenderId = senderId })
                 .ConfigureAwait(false);
 
-            return result == 1;
+            return transaction == null;
+        }
+
+        public async Task<bool> StoreTransactionIdsForSenderAsync(
+            string senderId,
+            IReadOnlyList<string> transactionIds,
```
lasrinnil, line 50: Is the transactionId collection checked for dups im memory?

source/Infrastructure/IncomingMessages/TransactionIdRegistry.cs
```diff
@@ -30,18 +34,56 @@ public TransactionIdRegistry(IDatabaseConnectionFactory connectionFactory)
...
+                    command.Parameters.Add(
+                        "@SenderId", SqlDbType.NChar).Value = senderId;
+                    await command.ExecuteScalarAsync(cancellationToken).ConfigureAwait(false);
+                }
+
+                await transaction.CommitAsync(cancellationToken).ConfigureAwait(false);
+            }
+            catch (SqlException)
+            {
+                await transaction.RollbackAsync(cancellationToken).ConfigureAwait(false);
```
lasrinnil, line 82: I cant seem to remember how behavior is expected by the business, if we enounter on transaction id which have been used in a series of transactions, should all transactions then be rejected?

source/Infrastructure.CimMessageAdapter/Messages/ITransactionIds.cs
```diff
@@ -23,11 +24,22 @@ namespace CimMessageAdapter.Messages
...
         /// <param name="senderId"></param>
         /// <param name="transactionId"></param>
         /// <param name="cancellationToken"></param>
-        Task<bool> TryStoreAsync(string senderId, string transactionId, CancellationToken cancellationToken);
+        Task<bool> TransactionIdOfSenderIsUniqueAsync(string senderId, string transactionId, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Store transaction ids for the specified sender
+        /// </summary>
+        /// <param name="senderId"></param>
```
lasrinnil, line 37: nit: remove unused params

source/Infrastructure.CimMessageAdapter/Messages/MessageReceiver.cs
```diff
@@ -89,53 +93,109 @@ public abstract class MessageReceiver<TQueue>
...
 
-                await AddToTransactionQueueAsync(transaction, cancellationToken).ConfigureAwait(false);
+            if (_errors.Count > 0)
+            {
+                return Result.Failure(_errors.ToArray());
+            }
+
+            try
+            {
+                using var scope = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled);
```
lasrinnil, line 120: Will review last of the PR tomorrow, too tired to be able to take in all these changes :D

source/Infrastructure/IncomingMessages/TransactionIdRegistry.cs
RasmusGraabaek: PO confirms that this is the intended behavior

source/Infrastructure/IncomingMessages/MessageQueueDispatcher.cs
MadsDue: We can see why this is need :)

MadsDue (approved):

PR #456, opened 2023-09-18
chore: Service bus commit fails test

RasmusGraabaek:
## Description

Renaming of some test and removed a service bus stub, which is not used anymore

## References

* #0000


source/IntegrationTests/Infrastructure.CimMessageAdapter/Stubs/MessageQueueDispatcherThatFailsStub.cs
```diff
@@ -1,62 +0,0 @@
...
-using System.Collections.Generic;
-using System.Threading;
-using System.Threading.Tasks;
-using Application.IncomingMessages;
-using CimMessageAdapter.Messages;
-using CimMessageAdapter.Messages.Queues;
-
-namespace IntegrationTests.Infrastructure.CimMessageAdapter.Stubs
-{
-    public class MessageQueueDispatcherThatFailsStub<TQueue> : MessageQueueDispatcherStub<TQueue>, IMessageQueueDispatcher<TQueue>
```
lasrinnil, line 25: Perhaps we should discuss why we are removing this instead of having a test that is dependent on it, that way we could have a test that verifies the expected behavior if a MessageQueueDispatcher fails when sending the ServiceBusMessages.

lasrinnil (approved):

PR #463, opened 2023-09-20
refac: major refactoring

lasrinnil:
## References

* #0000


source/Application/ArchivedMessages/GetArchivedMessageDocumentQueryHandler.cs
```diff
@@ -34,6 +34,6 @@ public GetArchivedMessageDocumentQueryHandler(IArchivedMessageRepository archive
     {
         ArgumentNullException.ThrowIfNull(request);
 
-        return await _archivedMessageRepository.GetDocumentAsync(request.Id, cancellationToken).ConfigureAwait(false);
+        return await _archivedMessageRepository.GetAsync(request.Id, cancellationToken).ConfigureAwait(false);
```
MadsDue, line 37: It gets the document and not the archivedMessage as I remember

PR #487, opened 2023-09-25
fix: Reduced aggregation bundle size to 1

RasmusGraabaek:
## Description
Currently bundles containing aggregations may contain 6 different messages. This is reduced to 1 to ensure we do not get "out of memory" exceptions


MadsDue (approved):
LGTM

PR #502, opened 2023-09-27
refac: message receiver to be a market message validator

MadsDue:
## Description
MessageReceiver was now only responsible for executing the sync validation rules. 
- [x] Rename MessageReceiver to MessageValidator
- [x] Since we only have 1 message to validat on it should accept this concrete type as a parameter.
- [x] Adjust InitializeAggregatedMeasureDataProcessesCommand to get RequestAggregatedMeasureDataMarketMessage
- [x] Adjust test to RequestAggregatedMeasureDataMarketDocumentBuilder for tests to build a RequestAggregatedMeasureDataMarketMessage

Next: https://github.com/Energinet-DataHub/opengeh-edi/pull/504
## References
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-phoenix/75


source/Domain/Documents/MarketMessage.cs
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+using Energinet.DataHub.EDI.Domain.Actors;
+
+namespace Energinet.DataHub.EDI.Domain.Documents;
+
+/// <summary>
+/// Represents a base market message.
+/// </summary>
+public record MarketMessage(
```
lasrinnil, line 22: Doesn't this represent the "market message" for a request for aggregated measured data?

source/Domain/Documents/RequestAggregatedMeasureDataMarketDocument.cs
```diff
@@ -12,20 +12,30 @@
...
+    string MessageId,
+    IReadOnlyCollection<Serie> Series) : MarketMessage(SenderNumber, SenderRole, ReceiverNumber, ReceiverRole, BusinessReason, AuthenticatedUser, AuthenticatedUserRole, MessageType, MessageId, Series);
 
 public record Serie(
     string Id,
     string? MarketEvaluationPointType,
     string? MarketEvaluationSettlementMethod,
-    string StartDateAndOrTimeDateTime,
-    string? EndDateAndOrTimeDateTime,
+    Instant StartDateAndOrTimeDateTime,
```
lasrinnil, line 37: why change this to another type? I thought we were taking a more leint approach since it will be Wholesales who is responsible for validation in most regards.

source/Domain/Documents/MarketMessage.cs
MadsDue: It was to keep the sync validator running on a generic type then the concrete aggregated measure data. But there is no need to do that now. So I  will remove this again. Good catch

PR #503, opened 2023-09-27
refac: Bump App package

dstenroejl:
I could not find a title prefix that matches "updating dependency", so I chose "refac:".

- [x] Update to latest App package to get "live" health checks that returns PR number and SHA commit.

Part of:
https://app.zenhub.com/workspaces/the-outlaws-6193fe815d79fc0011e741b1/issues/gh/energinet-datahub/team-the-outlaws/1435


MadsDue (approved):
Based on the story. I assume there is no breaking changes we should be aware about :)?

PR #539, opened 2023-10-03
fix: Updated settlement method to match documentation 

RasmusGraabaek:
## Description
The mapping of the settlement method to time series type did not match the documentation.


lasrinnil (approved):
@RSIEnerginet should probably verify that this is correct

PR #580, opened 2023-10-19
feat: Changed the eBix documentwriter so that it writes its on Messag…

jbn72:
…eContainer when returning a SOAP message


lasrinnil (commented):
Looks good! Would be great if there were a test that would ensure that "MessageContainer" and the other introduced elements are present when writing a Ebix document

jbn72:
Extended the validation of the ebix document structure to check for ebix message headers

lasrinnil (approved):

PR #587, opened 2023-10-23
refac: When receiving peeks in the application/ebix format we just fe…

jbn72:
…tches the oldest bundle from the DB


source/Domain/OutgoingMessages/Queueing/Bundle.cs
```diff
@@ -26,12 +26,13 @@ private Bundle()
     {
     }
 
-    internal Bundle(BundleId id, BusinessReason businessReason, DocumentType documentTypeInBundle, int maxNumberOfMessagesInABundle)
+    internal Bundle(BundleId id, BusinessReason businessReason, DocumentType documentTypeInBundle, int maxNumberOfMessagesInABundle, DateTime createdTS)
```
lasrinnil, line 29: Consider that the `createdTS `is simply set upon construction.

If the parameter is exposed for testing purposes, then please use a default of value of `null` and check for null upon construction, and if null then set the expected default value of `UtcNow.`

jbn72: Another way could be to set the value as a default value of the column in the database instead of doing it in code. The value is only used to find the "oldest" bundle to serve to the request

source/Domain/OutgoingMessages/Queueing/ActorMessageQueue.cs
```diff
@@ -100,18 +101,18 @@ private void EnsureApplicable(OutgoingMessage outgoingMessage)
             && bundle.BusinessReason == businessReason);
     }
 
-    private Bundle CreateBundleOf(BusinessReason businessReason, DocumentType messageType, int maxNumberOfMessagesInABundle)
+    private Bundle CreateBundleOf(BusinessReason businessReason, DocumentType messageType, int maxNumberOfMessagesInABundle, Instant timeStamp)
```
lasrinnil, line 104:
```suggestion
    private Bundle CreateBundleOf(BusinessReason businessReason, DocumentType messageType, int maxNumberOfMessagesInABundle, Instant created)
```

source/Tests/Domain/OutgoingMessages/Queueing/ActorMessageQueueTests.cs
```diff
@@ -123,8 +126,8 @@ public void Peek_returns_the_oldest_bundle()
         var actorMessageQueue = ActorMessageQueue.CreateFor(receiver);
         var messageAssignedToFirstBundle = CreateOutgoingMessage(receiver, BusinessReason.BalanceFixing);
         var messageAssignedToSecondBundle = CreateOutgoingMessage(receiver, BusinessReason.BalanceFixing);
-        actorMessageQueue.Enqueue(messageAssignedToFirstBundle, 1);
-        actorMessageQueue.Enqueue(messageAssignedToSecondBundle, 1);
+        actorMessageQueue.Enqueue(messageAssignedToFirstBundle, _now, 1);
+        actorMessageQueue.Enqueue(messageAssignedToSecondBundle, _now, 1);
```
lasrinnil, line 130: Perhaps we should create in a context where we have the database invovled to ensure that the "created" timestamp ensures that bundles stored in the DB are can be ordered and the oldest are peek'ed first.

jbn72: Parameter is resolved

jbn72: Test moved to other test project and is now testing using the DB (name "Can_peek_oldest_bundle"

source/Domain/OutgoingMessages/Queueing/ActorMessageQueue.cs
jbn72: Parameter is renamed

lasrinnil (approved):

PR #588, opened 2023-10-23
fix: wrong date format for b2c requests

MadsDue:
## Description
We should take instant format and not local datetime.

### DIFF
--- a/source/B2CWebApi/Factories/RequestAggregatedMeasureDataHttpFactory.cs
+++ b/source/B2CWebApi/Factories/RequestAggregatedMeasureDataHttpFactory.cs
@@ -81,7 +81,7 @@ public static class RequestAggregatedMeasureDataHttpFactory
 
     private static string SetTimeToMidnight(string dateString, DateTimeZone dateTimeZone)
     {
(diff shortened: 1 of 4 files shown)


source/B2CWebApi/Factories/RequestAggregatedMeasureDataHttpFactory.cs
```diff
@@ -81,10 +81,12 @@ public static class RequestAggregatedMeasureDataHttpFactory
 
     private static string SetTimeToMidnight(string dateString, DateTimeZone dateTimeZone)
     {
-        _date = InstantPattern.General.Parse(dateString).Value;
+        _date = InstantPattern.ExtendedIso.Parse(dateString).Value;
```
lasrinnil, line 84: Maybe a test to ensure that the expected time is the output?

MadsDue: Yes :) I will add it first thing tomorrow

RasmusGraabaek (approved):

PR #625, opened 2023-11-21
refac: Buildblocks infrastructure

lasrinnil:
## Description
To circumvent a circular reference from the `Infrastructure` project which holds our `UnitOfWork` that knows each `*.Infrastructure` to handle processing for each `DbContext` in a single transaction, and the fact that the shared infrastructure was placed in the same project. 

I have moved the `SQLConnetionFactory` and the `MessageBus` infrastructure stuff out and into a `BuildingBlocks.Infrastructure` project - I have been inspired to this approach from this repo: https://github.com/kgrzybek/modular-monolith-with-ddd

In the future "Common" will be empty (hopefully) and its content moved to corresponding `BuildingBlock` projects.
image


source/ArchivedMessages.Infrastructure/ArchivedMessages.Infrastructure.csproj
```diff
@@ -13,6 +13,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\ArchivedMessages.Interfaces\ArchivedMessages.Interfaces.csproj" />
```
MadsDue, line 15: Are Interfaces and Contracts representing the same thing or?

MadsDue, line 15: Should the Infrastructure layer know the Interfaces layer? should it not be the application.

MadsDue (approved):

source/ArchivedMessages.Infrastructure/ArchivedMessages.Infrastructure.csproj
lasrinnil: I'll do a refactor that will remove this dependency

lasrinnil: In another PR

PR #631, opened 2023-11-22
feat: use authenticated actor

MadsDue:


source/Api/OutgoingMessages/PeekRequestListener.cs
```diff
@@ -74,10 +74,10 @@ public class PeekRequestListener
         }
 
         var peekResult = await _mediator.Send(new PeekCommand(
-                _authenticator.CurrentIdentity.Number!,
-                msgCategory,
-                _authenticator.CurrentIdentity.Roles.First(),
-                desiredDocumentFormat)).ConfigureAwait(false);
+            _authenticatedActor.CurrentActorIdentity.ActorNumber,
+            msgCategory,
+            _authenticatedActor.CurrentActorIdentity.MarketRoles.First(),
```
lasrinnil, line 79: I believe there could only be one role pr. actor number

lasrinnil (approved):

PR #666, opened 2023-12-07
test: Test af MarketActorAuthenticatorMiddleware functions middleware

ebbeknudsen:
## Description
Test af MarketActorAuthenticatorMiddleware azure function middleware. FunctionContext er mock'd i `FunctionContextMock`, med de funktioner vi benytter i `MarketActorAuthenticatorMiddleware`


source/BuildingBlocks.Domain/Authentication/AuthenticatedActor.cs
```diff
@@ -38,7 +38,7 @@ public ActorIdentity CurrentActorIdentity
     /// <summary>
     /// Set the authenticated actor
     /// </summary>
-    public void SetAuthenticatedActor(ActorIdentity actorIdentity)
+    public void SetAuthenticatedActor(ActorIdentity? actorIdentity)
```
MadsDue, line 41: Instead of allowing us to set the authenticated actor as null, you should be able to do something like below, when you need to reset the authenticated actor for tests:
authenticatedActor.SetAuthenticatedActor(null!);

ebbeknudsen: Jeg synes umiddelbart at det er væsentligt pænere at vores kode kan tage imod null, hvis det er noget vi rent kan finde på at give den. Jeg tænker umiddelbart at det giver fint mening at man kan fjerne den authenticated user? Alternativt skulle være en separat funktion der hedder noget ala' `RemoveAuthenticatedActor()`

source/IntegrationTests/Api/Mocks/FunctionContextBuilder.cs
```diff
@@ -0,0 +1,53 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Security.Cryptography.X509Certificates;
+using Energinet.DataHub.Core.App.FunctionApp;
+
+namespace Energinet.DataHub.EDI.IntegrationTests.Api.Mocks;
+
+internal sealed class FunctionContextBuilder
```
lasrinnil, line 21: Consider adding a summary to describe what this is capable of - and its intended use.

lasrinnil: This comment would probably fit most of the classes in the/Mocks folder.

lasrinnil (approved):

PR #671, opened 2023-12-08
test: Add acceptance test for unauthorized requests

ebbeknudsen:
## Description

Add acceptance tests to ensure we do not allow access to our API without a bearer token


MadsDue (approved):
LGTM

PR #683, opened 2023-12-14
build: Increase tests build timeout period

ebbeknudsen:
## Description
Our test runs right now waits 3 minutes for the build to complete, then aborts. This PR increases it to 5 minutes. I think the only disadvantage on increasing the timeout is that if the build fails then the run will take longer to complete.

An alternative is to make the steps depend on the build, but then the tests overall time will be longer, since each test job won't start before the entire build is complete.

### DIFF
--- a/.github/workflows/ci-dotnet.yml
+++ b/.github/workflows/ci-dotnet.yml
@@ -27,27 +27,24 @@ jobs:
 
   # Run all tests in 'IntegrationTests.dll'
   integration_tests:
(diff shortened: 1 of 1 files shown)


MadsDue:
Why do we need to increase this :D?

MadsDue (approved):
LGTM

lasrinnil (approved):

Renetnielsen (approved):

PR #697, opened 2023-12-21
refac: MediatR out of IIntegrationEventProcessor

MWO1024:
## Description
Remove MediatR usage from the event processors.


source/Process.Application/IntegrationEvents/EnergyResultProducedV2Processor.cs
```diff
@@ -47,8 +51,8 @@ public async Task ProcessAsync(IntegrationEvent integrationEvent, CancellationTo
             .CreateAsync(energyResultProducedV2, CancellationToken.None)
             .ConfigureAwait(false);
 
-        var forwardAggregationResult = new ForwardAggregationResult(aggregation);
+        var message = AggregationResultMessageFactory.CreateMessage(aggregation, ProcessId.New());
 
-        await _mediator.Send(forwardAggregationResult, cancellationToken).ConfigureAwait(false);
```
ebbeknudsen, line 52: Shouldn't this just be `outgoingMessagesClient.EnqueueAsync()` and then you call `unitOfWork.CommitChangesAsync()` afterwards?

source/OutgoingMessages.Interfaces/IOutgoingMessagesClient.cs
```diff
@@ -42,4 +42,9 @@ public interface IOutgoingMessagesClient
     /// </summary>
     /// <param name="outgoingMessage"></param>
     Task EnqueueAsync(OutgoingMessageDto outgoingMessage);
+
+    /// <summary>
+    ///     Enqueue a message, WITH commit. Currently ONLY used by the Process module wrt reception of events.
+    /// </summary>
+    Task EnqueueAndCommitAsync(OutgoingMessageDto outgoingMessage, CancellationToken cancellationToken);
```
ebbeknudsen, line 49: As with our earlier talks, shouldn't this just be commit and then the caller uses `IUnitOfWork` to commit the changes afterwards?

source/ArchitectureTests/Application/InternalCommandTests.cs
```diff
@@ -28,7 +28,12 @@ public class InternalCommandTests
 {
     public static IEnumerable<object[]> GetInternalCommands()
     {
-        var allTypes = ApplicationAssemblies.Application.GetTypes().Concat(ApplicationAssemblies.Infrastructure.GetTypes()).Concat(typeof(ForwardAggregationResult).Assembly.GetTypes());
+        var allTypes = ApplicationAssemblies
+            .Application
+            .GetTypes()
+            .Concat(ApplicationAssemblies.Infrastructure.GetTypes())
+            .Concat(typeof(EnqueueMessageHandler).Assembly.GetTypes());
```
ebbeknudsen, line 35: This could maybe be refactored to something like `ApplicationAssemblies.ContainingInternalCommands`

source/OutgoingMessages.Interfaces/IOutgoingMessagesClient.cs
ebbeknudsen: Following up on this, shouldn't the commit call be moved up into our `IntegrationEventHandler`, or maybe even all the way up to our `IntegrationEventListener`?

What is our rule-of-thumb of when we call commit atm.?

ebbeknudsen (approved):

PR #715, opened 2024-01-08
Madu/update dependencies

MadsDue:


source/B2CWebApi/Security/FrontendUserProvider.cs
```diff
@@ -32,7 +32,7 @@ public FrontendUserProvider(AuthenticatedActor authenticatedActor)
     public Task<FrontendUser?> ProvideUserAsync(
         Guid userId,
         Guid actorId,
-        bool isFas,
+        bool multiTenancy,
```
lasrinnil, line 35: Is this change intended to be part of this PR?

lasrinnil (approved):

source/B2CWebApi/Security/FrontendUserProvider.cs
MadsDue: It was changed in the interface (from the nuget package) and stylecope like the names to be the same :D

PR #720, opened 2024-01-10
refac: Add the master data context to the unit of work

MWO1024:


MWO1024:
We ended up decided against this as there doesnt seem to be need for it at the moement, as the MDM isnt part of any longer workflow.

PR #726, opened 2024-01-12
refac: integrationevents should not be saved when processing fails

lasrinnil:
## Description
The previous implementation would always save an entry into the `ReceivedIntegrationEvents` table, even if the underlying processing would fail. This PR refactors the `ReceivedIntegrationEventRepository` to take an `IDBConnection` and  `IDBTransaction` to ensure that a rollback is possible if processing the event fails.


MadsDue (approved):
LGTM

PR #732, opened 2024-01-15
fix: accptance test to create new EnergyResultProducedV2 event

MadsDue:


source/Api/Program.cs
```diff
@@ -135,7 +136,7 @@ public static TokenValidationParameters DevelopmentTokenValidationParameters()
 
                     var integrationEventDescriptors = new List<MessageDescriptor>
                     {
-                        CalculationResultCompleted.Descriptor,
```
ebbeknudsen, line 138: Is this intentionel? AFAIK this is used for which integration events to receive in the `IntegrationEventHandler`

ebbeknudsen (approved):

PR #746, opened 2024-01-23
refac: Update and align actor roles

MWO1024:
## Description
The name(s) of the actor (market) roles are updated to align them across projects. Additionally, we change the persistence of actor roles to rely on the code (e.g. DGL) rather than the name.


source/Infrastructure/Configuration/IntegrationEvents/IntegrationEventMappers/ActorCertificateCredentialsAssignedEventProcessor.cs
```diff
@@ -60,18 +60,40 @@ private static MarketRole GetMarketRole(EicFunction actorRole)
             EicFunction.MeteringPointAdministrator => MarketRole.MeteringPointAdministrator,
             EicFunction.EnergySupplier => MarketRole.EnergySupplier,
             EicFunction.GridAccessProvider => MarketRole.GridOperator,
-            EicFunction.MeteredDataAdministrator => MarketRole.MeteringDataAdministrator,
+            EicFunction.MeteredDataAdministrator => MarketRole.MeteredDataAdministrator,
             EicFunction.MeteredDataResponsible => MarketRole.MeteredDataResponsible,
             EicFunction.BalanceResponsibleParty => MarketRole.BalanceResponsibleParty,
-            // TODO: MarketRole.CalculationResponsibleRole and MarketRole.MasterDataResponsibleRole cannot be created, since they are duplicates
-            _ => throw new ArgumentOutOfRangeException(nameof(actorRole), actorRole, "Unknown EicFunction market role value"),
+            EicFunction.ImbalanceSettlementResponsible => MarketRole.ImbalanceSettlementResponsible,
```
lasrinnil, line 66: We should not handle events with these EDI-unsupported `EicFunction`. So in the `ActorCertificateCredentialsAssignedEventProcessor` we should only process the supported `EicFunction`.

lasrinnil (approved):

lasrinnil (approved):

PR #748, opened 2024-01-23
feat: impl version number for request aggregated measure data

MadsDue:
## Description

- impl version for the calculation result and added it to the RSM-14 document if the value it present (not impl for EnergyResultProcceedV2 yet)

Next step, is to implement version as part of the result from wholesale in their repo + making the field not nullable

## References
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mandalorian/141


source/Tests/Infrastructure/OutgoingMessages/AggregationResult/AssertAggregationResultJsonDocument.cs
```diff
@@ -197,6 +198,12 @@ public IAssertAggregationResultDocument QualityIsNotPresentForPosition(int posit
         return this;
     }
 
+    public IAssertAggregationResultDocument HasCalculationResultVersion(int version)
+    {
+        Assert.Equal(version.ToString(NumberFormatInfo.InvariantInfo), FirstTimeSeriesElement().GetProperty("version").ToString());
```
MWO1024, line 203: Er der behov for et ekstra `.GetProperty("value")` her, eller slipper vi for det?

MadsDue: It is noy needed here. Because the "value" is on the version prop :)

PR #758, opened 2024-01-29
feat: only log ServiceBusTrigger calls as exception if its a retry.

lasrinnil:

### DIFF
--- a/source/Api/Configuration/ServiceBusExceptionLoggerMiddleware.cs
+++ b/source/Api/Configuration/ServiceBusExceptionLoggerMiddleware.cs
@@ -1,0 +1,65 @@
+﻿// Copyright 2020 Energinet DataHub A/S
+//
+// Licensed under the Apache License, Version 2.0 (the "License2");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


ebbeknudsen (approved):

source/Api/Configuration/ServiceBusExceptionLoggerMiddleware.cs
```diff
@@ -0,0 +1,65 @@
...
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
+
+    public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
+    {
+        ArgumentNullException.ThrowIfNull(context);
+        ArgumentNullException.ThrowIfNull(next);
+
+        // Check if the trigger is a Service Bus trigger
+        if (context.BindingContext.BindingData.ContainsKey("DeliveryCount"))
```
ebbeknudsen, line 38: We have an extension method on functioncontext called `Is()` that looks at triggertype, i don't know if it makes sense to use that one here instead?

ebbeknudsen (approved):

ebbeknudsen (approved):

PR #760, opened 2024-01-30
feat: different version for aggregated time series result

MadsDue:
## Description
aggregated time series now returns the latest calculation results for the requested period. 

This means that the result can be split into mulitple series elements (one per version). 

E.g.

over a period 01-01-2024 - 31-01-2024 there is 2 calculations: 

1. Calculation
```
Period 01-01-2024 - 31-01-2024
Version 1
```

2. Calculation
```
Period 10-01-2024 - 20-01-2024
Version 2
```

This will now return a series element per version:
```
Series : [
  {
    Period: 01-01-2024 - 09-01-2024
    Version: 1
    TimeSeriesPoint: [...]
  {
    Period: 10-01-2024 - 19-01-2024
    Version: 2
    TimeSeriesPoint: [...]
  },
  {
    Period: 20-01-2024 - 31-01-2024
    Version: 1
    TimeSeriesPoint: [...]
  }
]
```

## Story
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mandalorian/133

## References:
- https://github.com/Energinet-DataHub/opengeh-wholesale/pull/1829


source/Process.Domain/Transactions/AggregatedMeasureData/AggregatedMeasureDataProcess.cs
```diff
@@ -49,6 +51,8 @@ public class AggregatedMeasureDataProcess : Entity
             MeteringPointType = meteringPointType;
             SettlementMethod = settlementMethod;
             StartOfPeriod = startOfPeriod;
+            if (string.IsNullOrEmpty(endOfPeriod)) // Throw exception since our end period is nullable in our schema contract, but we validate for it in Wholesale
+                throw new ArgumentException("End of period shouldn't be able to be null, since validation in Wholesale rejects the request if it isn't set", nameof(endOfPeriod));
```
lasrinnil, line 55: Would this change existing behavior? If end is not supplied - should that end in a validation message for the actor to peek, or is it ok for them to be met with our generic error message?

source/Process.Domain/Transactions/AggregatedMeasureData/AggregatedMeasureDataProcess.cs
```diff
@@ -128,6 +132,17 @@ public void IsAccepted(IReadOnlyCollection<Aggregation> aggregations)
 
             if (_state == State.Sent)
             {
+                if (aggregations.Min(x => x.Period.Start) != InstantPattern.General.Parse(StartOfPeriod).Value
```
lasrinnil, line 135: If can't follow this expression in correspondence with the reject reason `_noDataAvailable`. Could you explain it to me?

source/Process.Domain/Transactions/Aggregations/OutgoingMessage/AggregationResultMessage.cs
```diff
@@ -64,7 +64,7 @@ public class AggregationResultMessage : OutgoingMessageDto
         Period period,
         IReadOnlyCollection<Point> points,
         string businessReasonName,
-        long? resultCalculationResultVersion,
+        long calculationResultVersion,
```
lasrinnil, line 67: Is this not breaking behavior? Is this chance intended for this PR?

source/Process.Domain/Transactions/AggregatedMeasureData/AggregatedMeasureDataProcess.cs
MadsDue: I found this check at another place and moved it in here, because the period is now coming from the serie element instead of the process. But lets us have a chat about it 😁

source/Process.Domain/Transactions/AggregatedMeasureData/AggregatedMeasureDataProcess.cs
```diff
@@ -21,11 +21,13 @@
 using Energinet.DataHub.EDI.Process.Domain.Transactions.AggregatedMeasureData.ProcessEvents;
 using Energinet.DataHub.EDI.Process.Domain.Transactions.Aggregations;
 using Energinet.DataHub.EDI.Process.Domain.Transactions.Aggregations.OutgoingMessage;
+using NodaTime.Text;
 
 namespace Energinet.DataHub.EDI.Process.Domain.Transactions.AggregatedMeasureData
 {
     public class AggregatedMeasureDataProcess : Entity
     {
+        private static readonly RejectReason _noDataAvailable = new("Ingen data tilgængelig / No data available", "E0H");
```
lasrinnil, line 30: This is part of the "async validation for incoming messages" I'm quite sure it was decided that all none sync validation would be performed by the subsystem in charge of the requested data (in this case Wholesale). Or am I just getting old?

source/RequestResponse/source/RequestResponse/Requests/AggregatedTimeSeriesRequestAccepted.proto
```diff
@@ -32,6 +32,8 @@ message Series {
    * The version of the calculation result. It is a positive integer (int64).
    */
   int64 calculation_result_version = 6;
+  google.protobuf.Timestamp start_of_period = 7;
```
lasrinnil, line 35: If these always should be used in unison, consider moving them unto a type that defines that.

MWO1024 (approved):

PR #759, opened 2024-01-29
feat: Market document in Azure Storage

ebbeknudsen:
## Description

- Change `MarketDocument` to use the related `ArchivedMessage` file storage reference, instead of duplicating the document file
- `IFileStorageClient` returns `DownloadedFile`
- `IArchivedMessagesClient` returns `IArchivedFile`
- Refactor `FileStorageReference.Category` to `FileStorageCategory` value object
- Change `DataLakeFileStorageClient` to stream the download, instead of downloading up front
- Add / update tests
- Clear related SQL tables (`ActorMessageQueues`, `Bundles`, and `MarketDocuments`)
- Change editorconfig rule instance fields with underscore to only apply to private fields
- Add `StreamValueObject` streams for incomming messages, archived messages & market document streams

## References
- https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/32

PR #771, opened 2024-02-05
refac: Removed unsed using

RasmusGraabaek:

### DIFF
--- a/source/ArchivedMessages.Interfaces/ArchivedMessage.cs
+++ b/source/ArchivedMessages.Interfaces/ArchivedMessage.cs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 using System;
-using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using Energinet.DataHub.EDI.BuildingBlocks.Domain.Models;
 using NodaTime;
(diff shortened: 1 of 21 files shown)


source/ArchivedMessages.Interfaces/ArchivedMessage.cs
```diff
@@ -43,7 +42,7 @@ public class ArchivedMessage
         BusinessReason = businessReason;
         Document = document;
 
-        var actorNumberForFileStorage = GetActorNumberForFileStoragePlacement(archivedMessageType, senderNumber, receiverNumber);
+        var actorNumberForFileStorage = GetActorNumberForFileStoragePlacement(archivedMessageType);
```
RasmusGraabaek, line 45: Am I messing something here?

ebbeknudsen: I changed it to use `senderNumber` and `receiverNumber` in PR #759

MadsDue (dismissed):
LGTM

source/ArchivedMessages.Interfaces/ArchivedMessage.cs
RasmusGraabaek: Cool, reverting

ebbeknudsen (approved):

PR #772, opened 2024-02-05
refac: move WholesaleInbox into Process moule

lasrinnil:
## Description

PR will move the `Wholesale inbox` with decencies into the Process module. I believe it is justifiable to do so based on the assumption EDI will for now only communicate with other sub-systems trough a process.

Besides just moving files - I have refactored a bit so `Process.Domain` would not have a depency for `ServiceBusMessage`'s

## References
There is no ref for this.


source/Process.Domain/Wholesale/IWholesaleInbox.cs
```diff
@@ -14,19 +14,19 @@
...
 {
     /// <summary>
-    /// Send <paramref name="request"/> to wholesale
+    /// Send <paramref name="aggregatedMeasureDataProcess"/> to wholesale
     /// </summary>
-    /// <param name="request"></param>
+    /// <param name="aggregatedMeasureDataProcess"></param>
     /// <param name="cancellationToken"></param>
-    Task SendAsync(ServiceBusMessage request, CancellationToken cancellationToken);
+    Task SendProcessAsync(AggregatedMeasureDataProcess aggregatedMeasureDataProcess, CancellationToken cancellationToken);
```
MadsDue, line 31: Before it could be used to send any messages as a ServiceBusMessage, now it should be a specific type?

source/Process.Infrastructure/Wholesale/WholesaleInbox.cs
```diff
@@ -16,31 +16,43 @@
...
-    public async Task SendAsync(
-        ServiceBusMessage request,
+    public async Task SendProcessAsync(
+        AggregatedMeasureDataProcess aggregatedMeasureDataProcess,
         CancellationToken cancellationToken)
     {
-        await _senderCreator.SendAsync(request, cancellationToken).ConfigureAwait(false);
+        ArgumentNullException.ThrowIfNull(aggregatedMeasureDataProcess);
+
+        var process = await _aggregatedMeasureDataProcessRepository
```
MadsDue, line 53: Is this process not the same as the one we get in the parameter, why do we need to load it again?
- I also like before where Send only was responsible for sending, not also loading and mapping :)

source/Process.Domain/Wholesale/IWholesaleInbox.cs
lasrinnil: Well - would ANY `ServicebusMessage` work? :)

source/Process.Infrastructure/Wholesale/WholesaleInbox.cs
lasrinnil: wooops! I had a commit I had not pushed, does this clear up the intent?

lasrinnil: **I also like before where Send only was responsible for sending, not also loading and mapping :)**

Loading is removed, the mapping is to keep the `ServiceBusMessage` dependencies out of `Process.Domain`. From a domain perspective I personally prefer a more specific approach to be explicit about the need to send a process to wholsale, rather than just any type of `ServiceBusMessage`. That sending it to wholesale is trough a `ServiceBusMessage` I view as a infrastructure concern.

source/Process.Infrastructure/Wholesale/WholesaleInbox.cs
```diff
@@ -37,10 +41,12 @@ public class WholesaleInbox : IWholesaleInbox
         _senderCreator = serviceBusSenderFactory.GetSender(options.Value.WHOLESALE_INBOX_MESSAGE_QUEUE_NAME);
     }
 
-    public async Task SendAsync(
-        ServiceBusMessage request,
+    public async Task SendProcessAsync(
+        AggregatedMeasureDataProcess aggregatedMeasureDataProcess,
```
RasmusGraabaek, line 45: Will this change be reverted, when we support requests of "månedssummer"?

lasrinnil: Perhaps ❓ 😄

RasmusGraabaek: Or are we going to have a send method for every type of process we have to send to wholesale?

MadsDue (approved):

source/Process.Infrastructure/InboxEvents/InboxEventsProcessor.cs
```diff
@@ -69,7 +69,9 @@ public async Task ProcessEventsAsync(CancellationToken cancellationToken)
             catch (Exception e)
             {
                 await MarkAsFailedAsync(inboxEvent, e, cancellationToken).ConfigureAwait(false);
+#pragma warning disable CA1727
```
RasmusGraabaek, line 72: needed?

PR #780, opened 2024-02-09
refac: remove unnecessary aggregation model

MadsDue:


lasrinnil (approved):

source/Process.Application/Transactions/Aggregations/AggregationMessageResultFactory.cs
```diff
@@ -20,132 +20,140 @@
...
-        if ((aggregatedMeasureDataProcess.MeteringPointType != null
-                ? MeteringPointType.FromCode(aggregatedMeasureDataProcess.MeteringPointType).Name
-                : null) != aggregatedTimeSerie.MeteringPointType)
-        {
-            throw new ArgumentException("AggregatedTimeSerie.MeteringPointType isn't equal to aggregatedMeasureDataProcess.MeteringPointType", nameof(aggregatedTimeSerie));
-        }
-
-        return new Aggregation(
-            MapPoints(aggregatedTimeSerie.Points),
+        var gridAreaDetails = await GetGridAreaDetailsAsync(aggregatedTimeSerie.GridAreaDetails.GridAreaCode, cancellationToken).ConfigureAwait(false);
```
RasmusGraabaek, line 55: Can we skip this step and use the gridAreaCode from aggregatedTImeSeries at line 60 instead?

source/Process.Application/Transactions/Aggregations/AggregationMessageResultFactory.cs
```diff
@@ -20,132 +20,140 @@
...
-    public async Task<Aggregation> CreateAsync(
+    public async Task<AggregationResultMessage> CreateAsync(
         EnergyResultProducedV2 integrationEvent,
         CancellationToken cancellationToken)
     {
         ArgumentNullException.ThrowIfNull(integrationEvent);
 
-        return new Aggregation(
-            MapPoints(integrationEvent.TimeSeriesPoints),
+        var gridAreaDetails = await GetGridAreaDetailsAsync(integrationEvent, cancellationToken).ConfigureAwait(false);
```
RasmusGraabaek, line 80: nit: Should we have a condition on this fetch, since line 85 indicates that we may fetch gridAreaDetails and not make use of it. 
I.e. if the message is to a energy supplier or balance responsible?

Or am I missing something?

RasmusGraabaek (changes requested):

source/Process.Application/Transactions/Aggregations/AggregationMessageResultFactory.cs
MadsDue: It is also use on line 88

source/Process.Application/Transactions/Aggregations/AggregationMessageResultFactory.cs
MadsDue: I think it is good for now. 

We need to do something with this GridAreaDetails in generel and now I just kept it similary to the mapping from EnergyResultProducedV2. 

I will bring it up in our dev meet up today :D

RasmusGraabaek (approved):

PR #783, opened 2024-02-13
chore: added retry strategy on EF db contexts

MadsDue:
## Description
Introduced  ResilientTransaction class to manage resilient database transactions using Entity Framework Core.
It uses the execution strategy from a DbContext instance to handle retries for transient failures.
If an action is specified, it will be executed before committing the transaction if action succeed.
E.g. to ensure database transaction is only committed if sending a message to a message broker success otherwise rollback.


source/IncomingMessages.Infrastructure/RequestAggregatedMeasureDataReceiver.cs
```diff
@@ -49,24 +49,32 @@ public class RequestAggregatedMeasureDataReceiver : IRequestAggregatedMeasureDat
...
 
-        using var transaction = await _incomingMessagesContext.Database.BeginTransactionAsync(cancellationToken)
-            .ConfigureAwait(false);
-
-        var result = await SaveMessageIdAndTransactionIdAsync(
-            requestAggregatedMeasureDataDto,
-            cancellationToken).ConfigureAwait(false);
-
-        if (result.Success)
+        // Use of an EF Core resiliency strategy
```
lasrinnil, line 52: personally I would perfer that the `ResilientTransaction` would have better documentation and these lines which appers multiple times were remove :)

source/Infrastructure/Configuration/DataAccess/ResilientTransaction.cs
```diff
@@ -0,0 +1,49 @@
...
+// limitations under the License.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Storage;
+
+namespace Energinet.DataHub.EDI.Infrastructure.Configuration.DataAccess;
+
+public class ResilientTransaction
```
lasrinnil, line 22: What is the intention for using this class? What does it help solve? Perhaps these questions could kickstart the documentation, please reach out to me if you want us to collaborate on it 💯

source/ApplyDBMigrationsApp/Scripts/Model/202402192140 New index on Bundles for ActorMessageQueue.sql
```diff
@@ -0,0 +1 @@
+﻿CREATE NONCLUSTERED INDEX IX_Bundles_ActorMessageQueue ON [dbo].[Bundles] ([ActorMessageQueueId]) INCLUDE ([BusinessReason], [Created], [DocumentTypeInBundle], [Id], [IsClosed], [IsDequeued], [MaxMessageCount], [MessageCount])
```
lasrinnil, line 1: This would have been nice if it was in its own PR. That way are more detailed explanation could go along with it (so I would understand the need for them) ;D

source/Infrastructure/Configuration/DataAccess/ResilientTransaction.cs
MadsDue: I have add some documentation to this class and moved it into building blocks.

source/BuildingBlocks.Infrastructure/DataAccess/ResilientTransaction.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Storage;
+
+namespace Energinet.DataHub.EDI.BuildingBlocks.Infrastructure.DataAccess;
+
+/// <summary>
+/// This responsible managing resilient database transactions using Entity Framework Core.
```
lasrinnil, line 23: I dont really get this line

lasrinnil (approved):

PR #787, opened 2024-02-17
feat: Initial steps to establish connection between archived messages

RasmusGraabaek:
## Description
We need to establish a connection between a request from an actor and the corresponding request.
Hence this pull request add a new field to store the id of the request.


source/OutgoingMessages.Domain/OutgoingMessages/Queueing/Bundle.cs
```diff
@@ -49,6 +50,9 @@ internal Bundle(BundleId id, BusinessReason businessReason, DocumentType documen
 
     public Instant Created { get; private set; }
 
+    // If this attribute is set, then the messages in the bundle are responses to a request from an actor
```
RasmusGraabaek, line 53: summary

source/OutgoingMessages.Domain/OutgoingMessages/Queueing/OutgoingMessage.cs
```diff
@@ -90,6 +91,8 @@ public OutgoingMessage(DocumentType documentType, ActorNumber receiverId, Guid p
 
         public FileStorageReference FileStorageReference { get; private set; }
 
+        public MessageId? RelatedToMessageId { get; set; }
```
RasmusGraabaek, line 94: Summary

source/OutgoingMessages.Interfaces/Models/OutgoingMessageDto.cs
```diff
@@ -57,4 +59,6 @@ public class OutgoingMessageDto
     public ActorRole SenderRole { get; }
 
     public string SerializedContent { get; }
+
+    public MessageId? RelatedToMessageId { get; }
```
RasmusGraabaek, line 63: summary

source/OutgoingMessages.Interfaces/Models/OutgoingMessageId.cs
```diff
@@ -16,6 +16,7 @@
 
 namespace Energinet.DataHub.EDI.OutgoingMessages.Interfaces.Models;
 
+// TODO: Is this just a messageId now?
```
RasmusGraabaek, line 19: !

source/OutgoingMessages.Infrastructure/OutgoingMessages/Queueing/ActorMessageQueueEntityConfiguration.cs
```diff
@@ -54,6 +54,8 @@ public void Configure(EntityTypeBuilder<ActorMessageQueue> builder)
             navigationBuilder.Property<int>("_messageCount").HasColumnName("MessageCount");
             navigationBuilder.Property<int>("_maxNumberOfMessagesInABundle").HasColumnName("MaxMessageCount");
             navigationBuilder.Property<Instant>("Created").HasColumnName("Created");
+            navigationBuilder.Property<MessageId>("RelatedToMessageId").HasColumnName("RelatedToMessageId")
```
ebbeknudsen, line 57: Can't we just use lambda's? Don't we only need to usedstrings if the properties are private:
`navigationBuilder.Property(p => p.RelatedToMessageId)`

This also exposes an error in the conversion, since `toDbValue` is nullable but we just use `toDbValue.Value` without checking first. So:
```c#
navigationBuilder.Property(p => p.RelatedToMessageId)
                .HasConversion(
                    toDbValue => toDbValue != null ? toDbValue.Value : null,
                    fromDbValue => fromDbValue != null ? MessageId.Create(fromDbValue) : null);
```

source/OutgoingMessages.Interfaces/Models/OutgoingMessageId.cs
ebbeknudsen: No :)

source/Process.Application/Transactions/AggregatedMeasureData/InitializeAggregatedMeasureDataProcessesHandler.cs
```diff
@@ -64,6 +64,7 @@ public class InitializeAggregatedMeasureDataProcessesHandler : IRequestHandler<I
                     actorSenderNumber,
                     dto.SenderRoleCode,
                     businessReason,
+                    MessageId.Create(dto.MessageId),
```
ebbeknudsen, line 67: Should probably move this outside the foreach like the `businessReason` to symbolize that it isn't coupled to the serie, but to the dto itself (and for a mini performance gain)

source/Process.Application/Transactions/Aggregations/AggregationMessageResultFactory.cs
```diff
@@ -58,7 +58,8 @@ public AggregationMessageResultFactory(IMasterDataClient masterDataClient)
             points: TimeSeriesPointsMapper.MapPoints(aggregatedTimeSerie.Points),
             businessReasonName: aggregatedMeasureDataProcess.BusinessReason.Name,
             calculationResultVersion: aggregatedTimeSerie.CalculationResultVersion,
-            settlementVersion: aggregatedMeasureDataProcess.SettlementVersion?.Name);
+            settlementVersion: aggregatedMeasureDataProcess.SettlementVersion?.Name,
+            relatedToMessageWithMessageId: aggregatedMeasureDataProcess.InitiatedByMessageId);
```
ebbeknudsen:
```suggestion
            relatedToMessageId: aggregatedMeasureDataProcess.InitiatedByMessageId);
```

ebbeknudsen (changes requested):

ebbeknudsen (dismissed):

ebbeknudsen (approved):

ebbeknudsen (approved):

PR #789, opened 2024-02-19
refac: ensure actors are informed when missing Content-Type header

lasrinnil:
## Description
When Actors peek they need to be informed when we are missing required headers, before this PR a 500 was returned if the content-type header was not set.

### DIFF
--- a/source/Api/OutgoingMessages/PeekRequestListener.cs
+++ b/source/Api/OutgoingMessages/PeekRequestListener.cs
@@ -58,12 +58,19 @@ public class PeekRequestListener
         ArgumentNullException.ThrowIfNull(request);
 
         var cancellationToken = request.GetCancellationToken(hostCancellationToken);
-        var contentType = request.Headers.GetContentType();
+        var contentType = request.Headers.TryGetContentType();
+        if (contentType is null)
+        {
+            _logger.LogInformation(
+                "Could not parse desired format from Content-Type header, support values are: application/xml, application/json, application/ebix");
+            return request.CreateResponse(HttpStatusCode.UnsupportedMediaType);
+        }
+
         var desiredDocumentFormat = DocumentFormatParser.ParseFromContentTypeHeaderValue(contentType);
         if (desiredDocumentFormat is null)
         {
             _logger.LogInformation(
-                "Could not parse desired CIM format from Content-Type header value: {ContentType}",
+                "Could not parse desired document format from Content-Type header value: {ContentType}",
                 contentType);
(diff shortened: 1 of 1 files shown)


source/Api/OutgoingMessages/PeekRequestListener.cs
```diff
@@ -58,12 +58,19 @@ public class PeekRequestListener
         ArgumentNullException.ThrowIfNull(request);
 
         var cancellationToken = request.GetCancellationToken(hostCancellationToken);
-        var contentType = request.Headers.GetContentType();
+        var contentType = request.Headers.TryGetContentType();
+        if (contentType is null)
+        {
+            _logger.LogInformation(
+                "Could not parse desired format from Content-Type header, support values are: application/xml, application/json, application/ebix");
```
MadsDue, line 65:
```suggestion
                "Could not get Content-Type from request header, support values are: application/xml, application/json, application/ebix");
```

MadsDue (approved):

source/Api/OutgoingMessages/PeekRequestListener.cs
```diff
@@ -58,12 +58,19 @@ public class PeekRequestListener
...
+            return request.CreateResponse(HttpStatusCode.UnsupportedMediaType);
+        }
+
         var desiredDocumentFormat = DocumentFormatParser.ParseFromContentTypeHeaderValue(contentType);
         if (desiredDocumentFormat is null)
         {
             _logger.LogInformation(
-                "Could not parse desired CIM format from Content-Type header value: {ContentType}",
+                "Could not parse desired document format from Content-Type header value: {ContentType}",
```
MadsDue, line 73:
```suggestion
                "Could not parse desired document format from Content-Type header value: {ContentType}, supported values are: application/xml, application/json, application/ebix",
```

source/Api/OutgoingMessages/PeekRequestListener.cs
```diff
@@ -58,12 +58,19 @@ public class PeekRequestListener
         ArgumentNullException.ThrowIfNull(request);
 
         var cancellationToken = request.GetCancellationToken(hostCancellationToken);
-        var contentType = request.Headers.GetContentType();
+        var contentType = request.Headers.TryGetContentType();
```
MadsDue, line 61: Can GetContentType() be removed now as well?

lasrinnil: IT is still in use in the request for aggregated measured data endpoint, ill make a PR to better handling the header content there as well :)

PR #790, opened 2024-02-19
refac: avoid return 500 errors for invalid requests

MadsDue:
## Description

This PR contains two changes: 
- missing content-type in request header is returned as 415
- aligned System.Text.Json version with explicit package referencing to ensure we catch excepted Json Parser exception (Unhandled JsonReaderException): https://stackoverflow.com/questions/34234433/unable-to-catch-jsonreaderexception


source/Common/Common.csproj
```diff
@@ -8,6 +8,7 @@
       <PackageReference Include="MediatR.Contracts" Version="2.0.1" />
       <PackageReference Include="NodaTime" Version="3.1.9" />
       <PackageReference Include="NodaTime.Serialization.SystemTextJson" Version="1.1.2" />
+      <PackageReference Include="System.Text.Json" Version="6.0.9" />
```
lasrinnil, line 11: Is the package really needed in every project it is referenced in? And if that is the case have you considered add'ing it to the `Directory.Build.props`

source/Api/IncomingMessages/RequestAggregatedMeasureMessageReceiver.cs
```diff
@@ -54,12 +54,20 @@ public class RequestAggregatedMeasureMessageReceiver
...
+            return request.CreateResponse(HttpStatusCode.UnsupportedMediaType);
+        }
+
         var documentFormat = DocumentFormatParser.ParseFromContentTypeHeaderValue(contentType);
         if (documentFormat is null)
         {
             _logger.LogInformation(
-                "Could not parse desired document format from Content-Type header value: {ContentType}",
+                "Could not parse desired document format from Content-Type header value: {ContentType}," +
+                "supported values are: application/xml, application/json, application/ebix",
```
RasmusGraabaek, line 70: Is: "supported values are: application/xml, application/json, application/ebix" something we want to find in our logs?

RasmusGraabaek: + Are documentation updated?

MadsDue: It was just to align it with this pr: https://github.com/Energinet-DataHub/opengeh-edi/pull/789/files

source/Common/Common.csproj
MadsDue: I am not sure that we need this in all projects going forward. I could see that we had it implicited references in all the update projects with different versions.

PR #791, opened 2024-02-20
feat: Support for related messages when searching for archived messages

RasmusGraabaek:


source/ArchivedMessages.Infrastructure/QueryBuilder.cs
```diff
@@ -87,19 +87,28 @@ public QueryInput BuildFrom(GetMessagesQuery request)
             throw new InvalidRestrictionException($"Invalid restriction for fetching archived messages. Must be either {nameof(Restriction.Owned)} or {nameof(Restriction.None)}. ActorNumber: {_actorIdentity.ActorNumber.Value}; Restriction: {_actorIdentity.Restriction.Name}");
         }
 
-        return new QueryInput(BuildStatement(), _queryParameters);
+        return new QueryInput(BuildStatement(request.IncludeRelatedMessage, request.MessageId), _queryParameters);
     }
 
-    private string BuildStatement()
+    public string BuildStatement(bool includeRelatedMessage, string? messageId = null)
```
ebbeknudsen, line 93: I don't think the `messageId = null` default value for is used anywhere. And i dont think the method needs to be public either.

source/ArchivedMessages.Infrastructure/QueryBuilder.cs
```diff
@@ -87,19 +87,28 @@ public QueryInput BuildFrom(GetMessagesQuery request)
...
     {
-        var selectStatement = "SELECT Id, MessageId, DocumentType, SenderNumber, ReceiverNumber, CreatedAt, BusinessReason FROM dbo.ArchivedMessages";
+        var whereClause = _statement.Count > 0 ? $" WHERE {string.Join(" AND ", _statement)}" : string.Empty;
+        string sqlStatement;
 
-        if (_statement.Count > 0)
+        if (includeRelatedMessage == true && messageId is not null)
         {
-            selectStatement += " WHERE " + string.Join(" AND ", _statement);
+            sqlStatement = "SELECT DISTINCT t2.Id, t2.MessageId, t2.DocumentType, t2.SenderNumber, t2.ReceiverNumber, t2.CreatedAt, t2.BusinessReason " +
```
ebbeknudsen, line 100: Maybe we should add a comment that explains what this sql statement does, and which edge cases it solves

source/IntegrationTests/Application/SearchMessages/SearchMessagesTests.cs
```diff
@@ -202,6 +202,76 @@ public async Task Filter_messages_by_business_reasons()
...
+        await ArchiveMessage(archivedMessage2);
+        await ArchiveMessage(archivedMessage21);
+
+        var messageIdOfMessage3 = MessageId.New();
+        var archivedMessage3 = CreateArchivedMessage(_systemDateTimeProvider.Now(), messageId: messageIdOfMessage3.Value);
+        await ArchiveMessage(archivedMessage3);
+
+        // Act
+        // This could simulate a search for a message, where it is a response to a request with more than one response
+        var resultForMessageId1 = await _archivedMessagesClient.SearchAsync(
```
ebbeknudsen, line 230: I think we should split these 3 searches into 3 seperate tests (so there is only 1 "act" / search in each test). I would probably keep the same data in each test

source/IntegrationTests/Application/SearchMessages/SearchMessagesTests.cs
```diff
@@ -202,6 +202,76 @@ public async Task Filter_messages_by_business_reasons()
...
+
+        // This could simulate a search for a message, where it is a request or response, but has no relation to any other messages
+        var resultForMessageId3 = await _archivedMessagesClient.SearchAsync(
+            new GetMessagesQuery(
+                MessageId: messageIdOfMessage3.Value,
+                IncludeRelatedMessage: true),
+            CancellationToken.None);
+
+        // Assert
+        Assert.Equal(3, resultForMessageId1.Messages.Count);
```
ebbeknudsen, line 251: We should assert that it is the correct messages that are returned, instead of that the count is correct (correct message id)

source/IntegrationTests/Application/SearchMessages/SearchMessagesTests.cs
```diff
@@ -202,6 +202,76 @@ public async Task Filter_messages_by_business_reasons()
...
+        await ArchiveMessage(archivedMessage1);
+        await ArchiveMessage(archivedMessage2);
+
+        var resultForMessageId = await _archivedMessagesClient.SearchAsync(
+            new GetMessagesQuery(
+                MessageId: messageId.Value,
+                IncludeRelatedMessage: false),
+            CancellationToken.None);
+
+        Assert.Single(resultForMessageId.Messages);
```
ebbeknudsen, line 272: Should assert that the returned message is the correct one (correct message id)

ebbeknudsen (changes requested):

source/IntegrationTests/Application/SearchMessages/SearchMessagesTests.cs
```diff
@@ -202,6 +203,140 @@ public async Task Filter_messages_by_business_reasons()
...
+                IncludeRelatedMessages: true),
+            CancellationToken.None);
+
+        // Assert
+        resultForMessageId3.Messages.Should().HaveCount(3);
+        resultForMessageId3.Messages.Should()
+            .OnlyContain(
+                message => message.MessageId == archivedMessage3.MessageId
+                           || message.MessageId == archivedMessage31.MessageId
+                           || message.MessageId == archivedMessage32.MessageId);
```
ebbeknudsen, line 244: Maybe something like:
```c#
using var scope = new AssertionScope();
resultForMessageId3.Messages.Should().HaveCount(3);

resultForMessageId3.Messages.Select(m => m.MessageId)
    .Should()
    .BeEquivalentTo(
        new[]
        {
            archivedMessage3.MessageId,
            archivedMessage31.MessageId,
            archivedMessage32.MessageId,
        });
```

The other statement could in theory be 3 result messages with the same message id's (ie. 3 result messages with the `archivedMessage31.MessageId` id)

source/IntegrationTests/Application/SearchMessages/SearchMessagesTests.cs
```diff
@@ -202,6 +203,140 @@ public async Task Filter_messages_by_business_reasons()
...
+                MessageId: messageIdOfMessage2.Value,
+                IncludeRelatedMessages: true),
+            CancellationToken.None);
+
+        // Assert
+        resultForMessageId2.Messages.Should().HaveCount(2);
+        resultForMessageId2.Messages.Should()
+            .OnlyContain(
+                message => message.MessageId == archivedMessage2.MessageId
+                           || message.MessageId == archivedMessage21.MessageId);
```
ebbeknudsen, line 283: ^ same as above (`.Should().BeEquivalentTo(...)`)

ebbeknudsen (approved):

PR #793, opened 2024-02-20
refac: avoid return 500 errors when content-type is missing requests

MadsDue:
## Description
When content-type is missing in request header, we are throwing a 500 error message to the actors without any error message. Instead we are now returning a 415 error telling them that the content-type is missing.


source/Api/IncomingMessages/RequestAggregatedMeasureMessageReceiver.cs
```diff
@@ -53,13 +53,20 @@ public class RequestAggregatedMeasureMessageReceiver
         ArgumentNullException.ThrowIfNull(request);
 
         var cancellationToken = request.GetCancellationToken(hostCancellationToken);
+        var contentType = request.Headers.TryGetContentType();
+        if (contentType is null)
+        {
+            _logger.LogInformation(
+                "Could not get Content-Type from request header.");
```
lasrinnil, line 60: why not leave them with a message that gives the solution? This way they could provide wrong information and would have to send another request to be informed of that

lasrinnil (approved):

PR #795, opened 2024-02-20
feat: Add CIM JSON for the-new-event-whats-it-called

MWO1024:

PR #805, opened 2024-02-23
refac: Fetch the CIM actor ids in acceptance tests

MWO1024:
## Description

Fetch the CIM actor ids instead on using hardcoded values. This should give us more flexibility in the future as we no longer require static actor numbers.

## References
#795


MWO1024:
Cant be done in the way we hoped. Maybe some other time.

PR #812, opened 2024-02-26
test: acceptance tests for amount per charge result

MadsDue:

### DIFF
--- a/source/AcceptanceTests/Drivers/MessageFactories/AmountPerChargeResultProducedV1Factory.cs
+++ b/source/AcceptanceTests/Drivers/MessageFactories/AmountPerChargeResultProducedV1Factory.cs
@@ -1,0 +1,60 @@
+﻿// Copyright 2020 Energinet DataHub A/S
+//
+// Licensed under the Apache License, Version 2.0 (the "License2");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Energinet.DataHub.Wholesale.Contracts.IntegrationEvents;
+using Energinet.DataHub.Wholesale.Contracts.IntegrationEvents.Common;
+using Google.Protobuf.Collections;
+using Google.Protobuf.WellKnownTypes;
+
+namespace Energinet.DataHub.EDI.AcceptanceTests.Drivers.MessageFactories;
+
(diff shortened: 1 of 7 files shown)


MWO1024 (approved):

source/AcceptanceTests/Drivers/MessageFactories/AmountPerChargeResultProducedV1Factory.cs
```diff
@@ -0,0 +1,61 @@
...
+            MeteringPointType = AmountPerChargeResultProducedV1.Types.MeteringPointType.Production,
+            CalculationResultVersion = 42,
+        };
+
+        amountPerChargeResultProduced.TimeSeriesPoints.Add(new RepeatedField<AmountPerChargeResultProducedV1.Types.TimeSeriesPoint>()
+        {
+            new AmountPerChargeResultProducedV1.Types.TimeSeriesPoint()
+            {
+                QuantityQualities = { AmountPerChargeResultProducedV1.Types.QuantityQuality.Calculated },
+                Quantity = new DecimalValue { Nanos = 1, Units = 0 },
```
RasmusGraabaek, line 52: Should we have add a point where non of these attributes has a value?

source/AcceptanceTests/Dsl/AggregationResultDsl.cs
```diff
@@ -53,4 +53,12 @@ internal async Task EmptyQueueForActor()
     {
         await _edi.EmptyQueueAsync().ConfigureAwait(false);
     }
+
+    internal Task PublishAmountPerChargeResultFor(
+        string gridAreaCode,
```
RasmusGraabaek, line 58: Maybe we should create a new DSL for this and "MonthlyAmountPerChargeResult", since these are "wholesaleResult"s?
Otherwise we could rename this class to, I dont know, "IntegretationEventDsl"?

lasrinnil: `EnergyResultDsl` and `WholesaleResultDsl` could be a way to distinguish between the two. But perhaps we should merge the two, as the both are representatives of a sum of a `WholesaleDsl`. What do you guys think?

source/AcceptanceTests/Drivers/MessageFactories/AmountPerChargeResultProducedV1Factory.cs
MadsDue: that is not the purpose of this test :D We have unit test that covers it

lasrinnil (commented):
Is the changes in this PR only to cover acceptance test for the enrgysupplier?

source/AcceptanceTests/Dsl/AggregationResultDsl.cs
MadsDue: I think we should have a WholesaleDsl responsible for pulishing the event etc. and a EdiDsl that is reponsible for comfiring the results.

RasmusGraabaek (approved):

RasmusGraabaek (approved):

PR #819, opened 2024-02-27
refac: B2C program is minimal hosting model

RasmusGraabaek:


source/B2CWebApi/Program.cs
```diff
@@ -12,24 +12,123 @@
...
+
+app.MapControllers().RequireAuthorization();
+
+app.MapLiveHealthChecks();
+
+app.MapReadyHealthChecks();
+
+app.Run();
+
+public partial class Program
```
MadsDue, line 132: Should we have a comment on why this is needed?

MadsDue (approved):
LGTM

PR #820, opened 2024-02-27
refac: Unit of work resolves the DbContexts

MWO1024:
## Description
We introduce a new marker class, for DbContexts that are to be included in the UoW. We then let UoW resolve these and commit all of them automatically.

## References
Follow-up to #813.


MWO1024:
This is a bit too complex. We will try something simpler.

PR #850, opened 2024-03-04
refac: outgoingMessage domain model is responsible for creating outgoing messages

MadsDue:
## Description
The changes in this Pull Request are primarily focused on the OutgoingMessage class, which is now responsible for correctly serialize various types to strings that later can become MarketActivityRecord objects and creating OutgoingMessages to be enqueued. 
This is tested in the OutgoingMessageTests.cs file.


source/Tests/Factories/EnergyResultMessageDtoBuilder.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            new(
+                1,
+                2,
+                CalculatedQuantityQuality.Calculated,
+                Instant.FromUtc(2024, 02, 02, 02, 02, 02).ToString()),
+        };
+
+    private static readonly string _businessReasonName = BusinessReason.BalanceFixing.Code;
+    private static readonly string? _originalTransactionIdReference = Guid.NewGuid().ToString();
```
RasmusGraabaek, line 52: This should not contain a value for an integration event

source/Tests/Factories/EnergyResultMessageDtoBuilder.cs
```diff
@@ -0,0 +1,77 @@
...
+    private static readonly string? _settlementVersion = SettlementVersion.FirstCorrection.Code;
+    private static readonly MessageId? _relatedToMessageId = MessageId.New();
```
RasmusGraabaek, line 54: This should not contain a value for an integration event

RasmusGraabaek (approved):

PR #857, opened 2024-03-04
refac: Remove composition root

RasmusGraabaek:


source/Process.Application/Extensions/DependencyInjection/ProcessExtensions.cs
```diff
@@ -43,9 +43,11 @@ public static IServiceCollection AddProcessModule(this IServiceCollection servic
...
                 "WHOLESALE_INBOX_MESSAGE_QUEUE_NAME must be set");
 
-        services.AddScopedSqlDbContext<ProcessContext>(configuration);
+        services
+            .AddScopedSqlDbContext<ProcessContext>(configuration)
+            .AddMediatR();
 
-        //EventsConfiguration
+        //InboxEventsConfiguration
         //TODO: can we move them out and delete ref to Infrastructure?
```
MWO1024, line 51: Infrastructure is dead, so I guess we should just remove this comment. Or maybe it is still highlightning a bad reference?

source/Process.Domain/Process.Domain.csproj
```diff
@@ -11,6 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="MediatR.Contracts" Version="2.0.1"/>
```
MWO1024, line 14: This makes me sad, but I don't think we can do anything about it.

source/Process.Infrastructure/Process.Infrastructure.csproj
```diff
@@ -15,6 +15,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\BuildingBlocks.Application\BuildingBlocks.Application.csproj"/>
```
MWO1024, line 18: Application reference in Infrastructure? 😨 This smells wrong

source/Process.Application/Extensions/DependencyInjection/ProcessExtensions.cs
RasmusGraabaek: They are located in process, so I'm deleting the comment

source/Process.Domain/Process.Domain.csproj
RasmusGraabaek: This is needed, since I moved DomainEvent to the process.domain solution
So where "DomainEvent"s go, mediatR.Contracts comes with it

source/Process.Infrastructure/Process.Infrastructure.csproj
RasmusGraabaek: Process' needs "ISystemDateTimeProvider" which are located in BuildingBlocks.application.

This may be wrong but it's also needed by our webApp

source/OutgoingMessages.Domain/OutgoingMessages.Domain.csproj
```diff
@@ -7,6 +7,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\BuildingBlocks.Domain\BuildingBlocks.Domain.csproj"/>
+    <ProjectReference Include="..\BuildingBlocks.Infrastructure\BuildingBlocks.Infrastructure.csproj"/>
```
RasmusGraabaek, line 10: If we want to avoid this, then outgoingMessages.domain has to have it's own serializer. Just like process.

Or we could move "Serializer" in application. And let "ISerializer" stay in infrastructure

MWO1024 (approved):

PR #868, opened 2024-03-08
feat: Enqueue NotifyAggregatedMeasureData for MDR in DDM queue

ebbeknudsen:
## Description
This feature is a "hack" that changes the outgoing message ActorMessageQueue for NotifyAggregatedMeasureData documents from MDR (MeteredDataResponsible) to DDM (GridOperator), since MDR/DDM actors has both roles, and always peeks the DDM queue, but makes RequestAggregatedMeasureData as a MDR

## References
Task: https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/145


source/OutgoingMessages.Domain/OutgoingMessages/Queueing/OutgoingMessage.cs
```diff
@@ -239,6 +239,22 @@ public string GetSerializedContent()
...
+        /// This is implemented to support the "hack" where a NotifyAggregatedMeasureData document for a MeteredDataResponsible
+        /// should be added to the GridOperator queue
+        /// </summary>
+        public Receiver GetActorMessageQueueMetadata()
+        {
+            var actorMessageQueueReceiverRole = ReceiverRole;
+
+            // NotifyAggregatedMeasureData document to MDR should always be added to the GridOperator queue
+            if (DocumentType == DocumentType.NotifyAggregatedMeasureData && ReceiverRole == ActorRole.MeteredDataResponsible)
+                actorMessageQueueReceiverRole = ActorRole.GridOperator;
```
MWO1024, line 253: Can't decide if I think

`            var actorMessageQueueReceiverRole =
                // NotifyAggregatedMeasureData document to MDR should always be added to the GridOperator queue
                DocumentType == DocumentType.NotifyAggregatedMeasureData
                && ReceiverRole == ActorRole.MeteredDataResponsible
                    ? ActorRole.GridOperator
                    : ReceiverRole;`

is more readable or not.

ebbeknudsen: The difference is small, but I think I prefer the if() statement since it reads as the flow diverging from the normal flow if this one specific case is happening

MWO1024: And it is easier to remove, if the hack is to be removed (we can hope).

MWO1024 (approved):

PR #869, opened 2024-03-11
feat: handler pattern for starting  new processes

MadsDue:
## Description
Prepare our Process module to handle multiple Initialize process types.


source/Api/EventListeners/ProcessInitializationListener.cs
```diff
@@ -26,30 +24,29 @@ namespace Energinet.DataHub.EDI.Api.EventListeners;
...
     }
 
     [Function(nameof(ProcessInitializationListener))]
     public async Task RunAsync(
         [ServiceBusTrigger(
             "%INCOMING_MESSAGES_QUEUE_NAME%",
             Connection = "SERVICE_BUS_CONNECTION_STRING_FOR_DOMAIN_RELAY_LISTENER")]
-        byte[] eventData,
+        ServiceBusReceivedMessage message,
         FunctionContext context)
```
RasmusGraabaek, line 43: I do believe that we can remove the context, since the message is typed

RasmusGraabaek (approved):

source/Process.Interfaces/IProcessInitializationClient.cs
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+using System.Threading.Tasks;
+
+namespace Energinet.DataHub.EDI.Process.Interfaces;
+
+/// <summary>
+/// Client for initializing processes.
+/// </summary>
+public interface IProcessInitializationClient
```
ebbeknudsen, line 22: Shouldn't this just be `IProcessClient` to follow the naming we use for our other clients?

ebbeknudsen (dismissed):

source/Process.Interfaces/IProcessInitializationClient.cs
MadsDue: Yes, you are rigth :D but currently, it is only used for initializing processes. But lets have a chat about it tomorrow :D

source/Api/EventListeners/ProcessInitializationListener.cs
```diff
@@ -26,30 +24,29 @@ namespace Energinet.DataHub.EDI.Api.EventListeners;
...
     {
+        ArgumentNullException.ThrowIfNull(message);
         ArgumentNullException.ThrowIfNull(context);
         var eventDetails = context.ExtractEventDetails();
```
RasmusGraabaek, line 47: When we hit this, the messageId of the service bus message does not exist. Hence the method "context.ExtractEventDetails();" til return an empty eventdetails. Giving us a useless log statement.

I would therefore prefer that we remove this loggin and might aswell kill the context from our trigger.

RasmusGraabaek (approved):

RasmusGraabaek (approved):

PR #895, opened 2024-03-18
test: ensure we always can write outgoing documents

MadsDue:
## Description
This PR adds testing to ensure we always can write the serilized message content for the market + a minor renaming


lasrinnil (approved):

source/Tests/Domain/OutgoingMessages/Queueing/OutgoingMessageTests.cs
```diff
@@ -44,8 +77,12 @@ public void Ensure_energy_result_can_be_deserialized_to_market_activity_record()
...
         // Assert
         var deserializedContent = serializer.Deserialize<TimeSeriesMarketActivityRecord>(outgoingMessage.GetSerializedContent());
-        energyResultMessageDto.Series.Should().BeEquivalentTo(deserializedContent);
-        energyResultMessageDto.Series.Point.Should().BeEquivalentTo(deserializedContent.Point);
+        energyResultMessageDto.Series.Should().BeEquivalentTo(
+            deserializedContent,
+            "because the point should be the same. If one is changed, the other should be changed as well. Remember to add the enqueuedQueuedSerializedContents with the new serialized content");
+        energyResultMessageDto.Series.Point.Should().BeEquivalentTo(
+            deserializedContent.Point,
+            "because the series point should be the same. If one is changed, the other should be changed as well. Remember to add the enqueuedQueuedSerializedContents with the new serialized content");
```
MWO1024, line 85: If I recall correctly, FluentAssertions adds the "because" to the error message itself, so it should be redundant here (and the same for the other tests).

MWO1024 (approved):

source/Tests/Domain/OutgoingMessages/Queueing/OutgoingMessageTests.cs
MadsDue: I dont believe it writes this part:  Remember to add the enqueuedQueuedSerializedContents with the new serialized content

PR #899, opened 2024-03-19
Science: Add PoC of new test layer

MWO1024:


MWO1024:
Just a PoC, so not needed.

PR #906, opened 2024-03-20
feat: use receiver on outgoing message to determind the actor queue

MadsDue:
## Description

When we are going to support delegation we need to be able to distinguish between the receiver in the message and which queue should receive the message

### DIFF
--- a/source/OutgoingMessages.Domain/OutgoingMessages/Queueing/OutgoingMessage.cs
+++ b/source/OutgoingMessages.Domain/OutgoingMessages/Queueing/OutgoingMessage.cs
@@ -28,6 +28,7 @@ namespace Energinet.DataHub.EDI.OutgoingMessages.Domain.OutgoingMessages.Queuein
         public static readonly FileStorageCategory FileStorageCategory = FileStorageCategory.OutgoingMessage();
 
         private string? _serializedContent;
+        private Receiver? _receiver;
 
         public OutgoingMessage(
             DocumentType documentType,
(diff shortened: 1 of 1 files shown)


source/OutgoingMessages.Domain/OutgoingMessages/Queueing/OutgoingMessage.cs
```diff
@@ -98,7 +99,14 @@ public class OutgoingMessage
...
 
-        public Receiver Receiver => Receiver.Create(ReceiverId, ReceiverRole);
+        /// <summary>
+        /// Receiver reference the actor queue that should receive the message.
+        /// </summary>
+        public Receiver Receiver
+        {
+            get => _receiver ?? Receiver.Create(ReceiverId, ReceiverRole);
+            set => _receiver = value;
+        }
```
ebbeknudsen, line 109: Maybe we should create a `DelegatedReceiver` instead of reusing the existing receiver? I think it's pretty confusing if the Receiver doesn't equal the `ReceiverId` and `ReceiverRole`

source/OutgoingMessages.Domain/OutgoingMessages/Queueing/OutgoingMessage.cs
```diff
@@ -291,7 +299,7 @@ public string GetSerializedContent()
         /// </summary>
         public Receiver GetActorMessageQueueMetadata()
         {
-            var actorMessageQueueReceiverRole = ReceiverRole;
+            var actorMessageQueueReceiverRole = Receiver.ActorRole;
```
ebbeknudsen, line 302: I think we should add a `DelegatedReceiver` and then this function uses the `DelegatedReceiver` instead of the original `Receiver`, if it is set

PR #909, opened 2024-03-20
feat: Change Wholesale requests to use names instead of CIM codes

ebbeknudsen:
## Description

- Add `EnumerationTypeWithCode<T>` base class for our domain value types
- Add `DomainNames` constants that are shared between EDI & Wholesale subsystems
- Change `AggregatedTimeSeriesRequest` to use domain names instead of CIM codes
- Change `WholesaleServicesRequest` to use domain names instead of CIM codes
- Test that `WholesaleServicesRequest` and `AggregatedTimeSeriesRequest` sent to Wholesale does not contain CIM codes


source/BuildingBlocks.Domain/DataHub/DomainNames.cs
```diff
@@ -0,0 +1,83 @@
...
+    public static class ChargeType
+    {
+        public const string Subscription = "Subscription";
+        public const string Fee = "Fee";
+        public const string Tariff = "Tariff";
+    }
+
+    public static class Currency
+    {
+        public const string DanishCrowns = "DanishCrowns";
```
RasmusGraabaek, line 47: I like "DanishCrowns", but I'm assuming that wholesale likes "dkk" more. But i do not know

source/BuildingBlocks.Domain/Models/BusinessReason.cs
```diff
@@ -13,19 +13,17 @@
...
+public sealed class BusinessReason : EnumerationTypeWithCode<BusinessReason>
 {
-    // ReSharper disable InconsistentNaming
-    #pragma warning disable IDE1006
-    public static readonly BusinessReason MoveIn = new(nameof(MoveIn), "E65");
-    public static readonly BusinessReason BalanceFixing = new(nameof(BalanceFixing), "D04");
-    public static readonly BusinessReason PreliminaryAggregation = new(nameof(PreliminaryAggregation), "D03");
-    public static readonly BusinessReason WholesaleFixing = new(nameof(WholesaleFixing), "D05");    //Engrosafiksering
-    public static readonly BusinessReason Correction = new(nameof(Correction), "D32");
+    public static readonly BusinessReason MoveIn = new(DomainNames.BusinessReason.MoveIn, "E65");
```
RasmusGraabaek, line 22: Some of our building blocks are serialized content.
Hence updating "DomainNmes" will be a breaking change, if one is not carefull

Which is like the current implementation :)

source/BuildingBlocks.Domain/Models/EnumerationTypeWithCode.cs
```diff
@@ -0,0 +1,67 @@
...
+               ?? throw new InvalidOperationException(
+                   $"{code} is not a valid {typeof(T).Name} {nameof(code)}");
+    }
+
+    public static T? TryFromCode(string code)
+    {
+        return GetAll<T>().FirstOrDefault(r => r.Code.Equals(code, StringComparison.OrdinalIgnoreCase));
+    }
+
+    public static string? GetNameFromCode(string code)
```
RasmusGraabaek, line 54: Is this also a "Try" method?

source/BuildingBlocks.Domain/Models/EnumerationType.cs
```diff
@@ -93,7 +93,7 @@ public override bool Equals(object? obj)
             }
 
             var typeMatches = GetType() == obj.GetType();
-            var valueMatches = Name.Equals(otherValue.Name, StringComparison.OrdinalIgnoreCase);
+            var valueMatches = ValueMatches(otherValue); // Name.Equals(otherValue.Name, StringComparison.OrdinalIgnoreCase);
```
RasmusGraabaek, line 96:
```suggestion
            var valueMatches = ValueMatches(otherValue);
```

source/IntegrationTests/FluentAssertionExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    /*
+     * This RegEx test if a string looks like a CIM code
+     * Here is an example: https://regex101.com/r/HhGUH4/1
+     * The regex thinks the string is a CIM code if:
+     * 1. The string MUST start with A, D, E, P, K, or H
+     * 2. The string MUST be followed by 2 characters
+     * 3. The string can be followed by 0 to 2 characters
+     * So in total: The string start with A, D, E, P, K, or H has a total length of 3 to 5 characters
+     * Match examples: D01, D32, E02, PT15M, P1D, P1M, KWH, H87, A09, E65, PT1H
```
RasmusGraabaek, line 32: Are our market roles comsidered CIM codes? aka MDR and so on?

RasmusGraabaek (dismissed):

source/BuildingBlocks.Domain/DataHub/DomainNames.cs
```diff
@@ -0,0 +1,83 @@
...
+
+/// <summary>
+/// These domain names are shared between the EDI and Wholesale subsystems
+/// When updating these, you need to manually update the classes in the other subsystem
+/// Files to manually keep in sync:
+/// - EDI: BuildingBlocks.Domain/DataHub/DomainNames.cs
+/// - Wholesale: Edi/Edi/Contracts/DomainNames.cs
+/// </summary>
+[SuppressMessage("Design", "CA1034:Nested types should not be visible", Justification = "Keep names in a single file to easily share with Wholesale")]
+public static class DomainNames
```
lasrinnil, line 27: I would really prefer if we ended up with sharing a common package with these "names" :) but its a good start!

source/BuildingBlocks.Domain/Models/ChargeType.cs
```diff
@@ -13,39 +13,19 @@
...
 
 using System;
-using System.Linq;
+using Energinet.DataHub.EDI.BuildingBlocks.Domain.DataHub;
 
 namespace Energinet.DataHub.EDI.BuildingBlocks.Domain.Models;
 
 [Serializable]
-public class ChargeType : EnumerationType
+public class ChargeType : EnumerationTypeWithCode<ChargeType>
```
lasrinnil, line 21:
```suggestion
public class ChargeType : DataHubType<ChargeType>
```

Please let me try to explain this concept tomorrow :)

source/IntegrationTests/FluentAssertionExtensions.cs
lasrinnil: They are not - lets all together try to define which is which

source/BuildingBlocks.Domain/Models/Resolution.cs
```diff
@@ -13,42 +13,27 @@
...
-    }
-
-    public static Resolution FromCode(string code)
-    {
-        return GetAll<Resolution>().FirstOrDefault(r => r.Code.Equals(code, StringComparison.OrdinalIgnoreCase))
-               ?? throw new InvalidOperationException($"{code} is not a valid {typeof(Resolution)} {nameof(code)}");
-    }
-
-    public static Resolution? TryFromCode(string code)
+    private static Resolution CreateUnknown(string? name, string? code)
```
lasrinnil, line 35: private / comment when to use

source/IntegrationTests/FluentAssertionExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Energinet.DataHub.EDI.IntegrationTests;
+
+public static class FluentAssertionExtensions
+{
+    /*
+     * This RegEx test if a string looks like a CIM code
+     * Here is an example: https://regex101.com/r/HhGUH4/2
+     * The regex thinks the string is a CIM code if:
+     * 1. The string MUST start with A, D, E, P, K, H, M, S
+     * 2. The string MUST be followed by 2 characters
```
lasrinnil, line 29: 2 breaks 3?

source/Process.Infrastructure/Transactions/WholesaleRequestNameConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Energinet.DataHub.EDI.BuildingBlocks.Domain.Models;
+
+namespace Energinet.DataHub.EDI.Process.Infrastructure.Transactions;
+
+public static class WholesaleRequestNameConverter
```
lasrinnil, line 19: What is the benefit of this class instead of calling ex. ActorRole.TryGetNameFromCode(actorRoleCode) ?? actorRoleCode; instead of ActorRoleCodeToName(string actorRoleCode)?

source/BuildingBlocks.Domain/Models/Resolution.cs
ebbeknudsen: It is unused so I've deleted it

source/IntegrationTests/FluentAssertionExtensions.cs
ebbeknudsen: No, it's in order. The string character must be followed by 2 characters (min length = 3), and then it can be followed by 0-2 characters (length = 3-5). I could merge step 2 and 3 in the description I guess, but it is based on how the regex is split

source/Process.Infrastructure/Transactions/WholesaleRequestNameConverter.cs
ebbeknudsen: Just to reuse / not copy that line between both `WholesaleServicesRequestFactory` and `AggregatedMeasureDataRequestFactory`

source/BuildingBlocks.Domain/Models/SettlementVersion.cs
```diff
@@ -13,17 +13,17 @@
...
 {
-    public static readonly SettlementVersion FirstCorrection = new(nameof(FirstCorrection), "D01");
-    public static readonly SettlementVersion SecondCorrection = new(nameof(SecondCorrection), "D02");
-    public static readonly SettlementVersion ThirdCorrection = new(nameof(ThirdCorrection), "D03");
+    public static readonly SettlementVersion FirstCorrection = new(DataHubNames.SettlementVersion.FirstCorrection, "D01");
+    public static readonly SettlementVersion SecondCorrection = new(DataHubNames.SettlementVersion.SecondCorrection, "D02");
+    public static readonly SettlementVersion ThirdCorrection = new(DataHubNames.SettlementVersion.ThirdCorrection, "D03");
 
     // Below SettlementVersions are not used directly, but must be here for possible mapping
     public static readonly SettlementVersion FourthCorrection = new(nameof(FourthCorrection), "D04");
```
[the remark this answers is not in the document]
ebbeknudsen: It's not one we are supporting or using anywhere in EDI / Wholesale atleast, but it is known by our schema (just like the 100 business reasons we aren't using)

source/Process.Infrastructure/Transactions/WholesaleRequestNameConverter.cs
ebbeknudsen: I guess it's only the `ActorRoleCodeToName` that is actually reused between the two factories, so they could just be written inside the actual factories. The method we use for converting the code to name is just similar between the different types, so I just bundled them together

ebbeknudsen: I've moved this implementation into the factories (and into the base DataHubType), and deleted the WholesaleRequestNameConverter

lasrinnil (approved):

PR #917, opened 2024-03-21
refac: more explixit about the outgoing message receiver and document receiver

MadsDue:

### DIFF
--- a/source/IntegrationTests/Application/OutgoingMessages/WhenEnqueueingOutgoingMessageTests.cs
+++ b/source/IntegrationTests/Application/OutgoingMessages/WhenEnqueueingOutgoingMessageTests.cs
@@ -90,15 +90,16 @@ public class WhenEnqueueingOutgoingMessageTests : TestBase
             () => Assert.NotNull(result!.RecordId),
             () => Assert.Equal(message.ProcessId, result!.ProcessId),
             () => Assert.Equal(DocumentType.NotifyAggregatedMeasureData.Name, result!.DocumentType),
-            () => Assert.Equal(message.ReceiverId.Value, result!.ReceiverId),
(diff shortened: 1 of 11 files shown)


MWO1024 (approved):

source/IntegrationTests/Application/OutgoingMessages/WhenEnqueueingOutgoingMessageTests.cs
```diff
@@ -90,15 +90,16 @@ public async Task Outgoing_message_is_added_to_database_with_correct_values()
             () => Assert.NotNull(result!.RecordId),
             () => Assert.Equal(message.ProcessId, result!.ProcessId),
             () => Assert.Equal(DocumentType.NotifyAggregatedMeasureData.Name, result!.DocumentType),
-            () => Assert.Equal(message.ReceiverId.Value, result!.ReceiverId),
+            () => Assert.Equal(message.ReceiverId.Value, result!.ReceiverNumber),
```
ebbeknudsen, line 93: Should we rename `ReceiverId` -> `ReceiverNumber` on the `OutgoingMessageDto` as well, or is that a bigger change / breaking change etc.?

ebbeknudsen (approved):

source/IntegrationTests/Application/OutgoingMessages/WhenEnqueueingOutgoingMessageTests.cs
MadsDue: No, let's do it. Good catch!

MadsDue: https://github.com/Energinet-DataHub/opengeh-edi/pull/918

PR #923, opened 2024-03-27
fix: ensure actor can request monthly wholesale service sum

MadsDue:


lasrinnil (approved):
No additional tests are required to very this change?

lasrinnil (approved):

PR #931, opened 2024-04-02
test: run acceptance tests against dev00x

MadsDue:
## References
- https://github.com/Energinet-DataHub/dh3-dev-secrets/pull/36

### DIFF
--- a/source/AcceptanceTests/AcceptanceTestFixture.cs
+++ b/source/AcceptanceTests/AcceptanceTestFixture.cs
@@ -41,7 +41,7 @@ public class AcceptanceTestFixture : IAsyncLifetime
     public AcceptanceTestFixture()
     {
         var configurationBuilder = new ConfigurationBuilder()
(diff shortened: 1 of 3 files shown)


source/AcceptanceTests/AcceptanceTestFixture.cs
```diff
@@ -41,7 +41,7 @@ public class AcceptanceTestFixture : IAsyncLifetime
     public AcceptanceTestFixture()
     {
         var configurationBuilder = new ConfigurationBuilder()
-            .AddJsonFile("acceptancetest.local.settings.json", true)
```
lasrinnil, line 44: is the `acceptancetest.local.settings.json` still usefull after this change?

lasrinnil (approved):

source/AcceptanceTests/AcceptanceTestFixture.cs
MadsDue: I don't think it would make sense to run the acceptance test locally :)

PR #943, opened 2024-04-05
fix: ensure we write quantity quality correct

MadsDue:

### DIFF
--- a/source/OutgoingMessages.Application/MarketDocuments/NotifyWholesaleServices/NotifyWholesaleServicesEbixDocumentWriter.cs
+++ b/source/OutgoingMessages.Application/MarketDocuments/NotifyWholesaleServices/NotifyWholesaleServicesEbixDocumentWriter.cs
@@ -190,11 +190,11 @@ public class NotifyWholesaleServicesEbixDocumentWriter : EbixDocumentWriter
                             }
 
                             // <QuantityQuality />
-                            if (point.QuantityQuality != null && EbixCode.WholesaleServicesOf(point.QuantityQuality.Value) is not null)
+                            if (point.QuantityQuality != null && EbixCode.ForWholesaleServicesOf(point.QuantityQuality.Value) is not null)
                             {
                                 await WriteCodeWithCodeListReferenceAttributesAsync(
                                         "QuantityQuality",
-                                        EbixCode.WholesaleServicesOf(point.QuantityQuality.Value)!,
+                                        EbixCode.ForWholesaleServicesOf(point.QuantityQuality.Value)!,
                                         writer)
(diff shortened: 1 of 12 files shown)


source/OutgoingMessages.Application/MarketDocuments/NotifyWholesaleServices/NotifyWholesaleServicesEbixDocumentWriter.cs
```diff
@@ -190,9 +190,13 @@ await writer.WriteElementStringAsync(DocumentDetails.Prefix, "Position", null, p
                             }
 
                             // <QuantityQuality />
-                            if (point.QuantityQuality != null)
+                            if (point.QuantityQuality != null && EbixCode.ForWholesaleServicesOf(point.QuantityQuality.Value) is not null)
```
RasmusGraabaek, line 193: A method, as you have done in json and xml would be nice  (WriteQualityIfSpecified)

source/Tests/Infrastructure/OutgoingMessages/Asserts/AssertXmlDocument.cs
```diff
@@ -87,6 +87,7 @@ public async Task<AssertXmlDocument> HasValidStructureAsync(DocumentType type, s
     {
         ArgumentNullException.ThrowIfNull(_documentValidator);
         var validationResult = await _documentValidator.ValidateAsync(_stream, DocumentFormat.Xml, type, CancellationToken.None, version).ConfigureAwait(false);
+        validationResult.ValidationErrors.Should().BeEmpty();
```
RasmusGraabaek, line 90: If I recall correct 
validationResult.ValidationErrors.Should().BeEmpty()
 is the same as 
validationResult.IsValid

Since validationResult.IsValid should be defined as ValidationErrors.Count == 0

I may be wrong

RasmusGraabaek (approved):

source/Tests/Infrastructure/OutgoingMessages/Asserts/AssertXmlDocument.cs
MadsDue: Correct, is it just easier to get the validation errors when debugging in this way :D

source/OutgoingMessages.Application/MarketDocuments/NotifyWholesaleServices/NotifyWholesaleServicesEbixDocumentWriter.cs
MadsDue: I will includes this in the next pr

PR #947, opened 2024-04-08
fix: Transaction id on processes should only be unique for the requested by actor

MadsDue:

### DIFF
--- a/source/ApplyDBMigrationsApp/Scripts/Model/202404081346 Update unique key constraint on BusinessTransactionId in AggregatedMeasureDataProcesses.sql
+++ b/source/ApplyDBMigrationsApp/Scripts/Model/202404081346 Update unique key constraint on BusinessTransactionId in AggregatedMeasureDataProcesses.sql
@@ -2,4 +2,4 @@
     DROP CONSTRAINT UC_BusinessTransactionId
          
 ALTER TABLE [dbo].[AggregatedMeasureDataProcesses]
-    ADD CONSTRAINT UC_BusinessTransactionIdAndRequestedByActorId UNIQUE (BusinessTransactionId, RequestedByActorId);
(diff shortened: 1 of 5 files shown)


source/ApplyDBMigrationsApp/Scripts/Model/202404081346 Update unique key constraint on BusinessTransactionId in AggregatedMeasureDataProcesses.sql
```diff
@@ -0,0 +1,5 @@
+﻿ALTER TABLE [dbo].[AggregatedMeasureDataProcesses]
+    DROP CONSTRAINT UC_BusinessTransactionId
+         
+ALTER TABLE [dbo].[AggregatedMeasureDataProcesses]
+    ADD CONSTRAINT UC_AggreBusinessTransactionIdAndRequestedByActorId UNIQUE (BusinessTransactionId, RequestedByActorId);
```
ebbeknudsen, line 5: I'm used to a naming convention of something like `UX(or UC?)_TableName_ColumnName1_ColumnName2_etc`, so `UX_AggregatedMeasureDataProcesses_BusinessTransactionId_RequestedByActorId` in this case. I don't know if we use any  convention in EDI so I'll let it be up to you :)

ebbeknudsen (dismissed):

source/ApplyDBMigrationsApp/Scripts/Model/202404081346 Update unique key constraint on BusinessTransactionId in AggregatedMeasureDataProcesses.sql
MadsDue: Don't think we have any. I will buy yours ;)

ebbeknudsen (approved):

PR #958, opened 2024-04-12
fix: Incorrect naming of behaviour test/helper method

MWO1024:

### DIFF
--- a/source/IntegrationTests/Behaviours/IncomingRequests/GivenAggregatedMeasureDataRequestTests.cs
+++ b/source/IntegrationTests/Behaviours/IncomingRequests/GivenAggregatedMeasureDataRequestTests.cs
@@ -81,7 +81,7 @@ public class GivenAggregatedMeasureDataRequestTests : BehavioursTestBase
 
     [Fact]
     public async Task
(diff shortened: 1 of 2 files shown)


source/IntegrationTests/Behaviours/IncomingRequests/GivenAggregatedMeasureDataRequestTests.cs
```diff
@@ -81,7 +81,7 @@ public async Task
 
     [Fact]
     public async Task
-        Given_RequestAggregatedMeasureDataJsonIsReceivedAndDelegation_When_ActorPeeksJson_Then_CorrectJsonDocumentIsCreated()
+        Given_RequestAggregatedMeasureDataJsonIsReceived_When_ActorPeeksJson_Then_CorrectJsonDocumentIsCreated()
```
ebbeknudsen, line 84: Remember that the test should be format agnostic, and thus receive the DocumentFormat as a method parameter using `[Theory]` and `[MemberData]`

ebbeknudsen (approved):

ebbeknudsen (approved):

PR #959, opened 2024-04-12
test: Add behaviour test for AmountPerChargeResultProducedV1

ebbeknudsen:
## Description
Add Proof-of-Concept for first integration test (GivenAmountPerChargeResultProducedV1ReceivedTests) using our new behaviour test structure


source/IntegrationTests/Behaviours/IntegrationEvents/GivenAmountPerChargeResultProducedV1ReceivedTests.cs
```diff
@@ -0,0 +1,125 @@
...
+                .WithSettlementMethod(AmountPerChargeResultProducedV1.Types.SettlementMethod.Flex)
+                .WithMeteringPointType(AmountPerChargeResultProducedV1.Types.MeteringPointType.Consumption)
+                .WithGridAreaCode("244")
+                .WithQuantityUnit(AmountPerChargeResultProducedV1.Types.QuantityUnit.Kwh)
+                .WithCalculationVersion(1)
+                .WithResolution(AmountPerChargeResultProducedV1.Types.Resolution.Hour)
+                .WithStartOfPeriod(CreateDateInstant(2022, 09, 06).ToTimestamp())
+                .WithEndOfPeriod(CreateDateInstant(2022, 09, 07).ToTimestamp())
+                .WithPoint(1, 10, 3, 30, AmountPerChargeResultProducedV1.Types.QuantityQuality.Calculated) // QuantityQualityCodeCalculated = "A06". "A01" from example doesn't exist?
+                .WithIsTax(false));
```
MWO1024, line 79: I think we should, at the very least, secide if we want to create events via builder pattern, or factory methods. Personally, I think factory methods is the way to go for these tests, to ensure/enforce that we think about the value of each field.

MWO1024: Not necessarily something we have to change now.

MWO1024 (approved):

source/IntegrationTests/Behaviours/IntegrationEvents/GivenAmountPerChargeResultProducedV1ReceivedTests.cs
```diff
@@ -0,0 +1,125 @@
...
+
+#pragma warning disable CS1570 // XML comment has badly formed XML
+    /// <summary>
+    /// Example based on: https://energinet.sharepoint.com/sites/DH3ART-team/_layouts/15/download.aspx?UniqueId=039914efcff74ec1a159eff3bb358f68&e=0RN8ma
+    /// </summary>
+    /// <param name="documentFormat"></param>
+#pragma warning restore CS1570 // XML comment has badly formed XML
+    [Theory]
+    [MemberData(nameof(AllDocumentFormats))]
+    public async Task When_EnergySupplierActorPeeksMessage_Then_ReceivesCorrectNotifyWholesaleServicesDocument(DocumentFormat documentFormat)
```
RasmusGraabaek, line 55: You can add [ExcludeFromNameConventionCheck]
to the test, if you want to have another name convention than "Given_x_When_y_Then_z"

source/IntegrationTests/Behaviours/IntegrationEvents/GivenAmountPerChargeResultProducedV1ReceivedTests.cs
ebbeknudsen: I think we should refactor to use a factory method that requires all known values is supplied, however i think we should do that in another PR 👍

source/IntegrationTests/Behaviours/IntegrationEvents/GivenAmountPerChargeResultProducedV1ReceivedTests.cs
ebbeknudsen: I've exented the regex test since i think it's a valid test name to have the "Given" statement be in the class name, and instead of repeating it in the method name then only write the "When" and "Then" statements

PR #961, opened 2024-04-15
feat: metered data responsible should also check delegation relations as grid operator

MadsDue:
We need to ensure that messages to metered data repsonsible is also delageted to a new receiver if there is a delegation relationship setup for the actor as a grid operator.

### DIFF
--- a/source/IntegrationTests/Application/OutgoingMessages/WhenEnqueueingOutgoingMessageWithDelegationTests.cs
+++ b/source/IntegrationTests/Application/OutgoingMessages/WhenEnqueueingOutgoingMessageWithDelegationTests.cs
@@ -55,6 +55,10 @@ public class WhenEnqueueingOutgoingMessageWithDelegationTests : TestBase
         _dateTimeProvider = (SystemDateTimeProviderStub)GetService<ISystemDateTimeProvider>();
     }
 
+    /// <summary>
+    /// This is implemented to support the "hack" where
+    ///     MeteredDataResponsible is working as GridOperator.
+    /// </summary>
     [Fact]
     public async Task
         Given_OutgoingEnergyResultMessageToMeteredDataResponsible_When_DelegatedByIsGridOperator_Then_GridOperatorReceivesMessage()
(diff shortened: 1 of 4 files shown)


source/IntegrationTests/Application/OutgoingMessages/WhenEnqueueingOutgoingMessageWithDelegationTests.cs
```diff
@@ -55,6 +55,31 @@ public WhenEnqueueingOutgoingMessageWithDelegationTests(IntegrationTestFixture i
         _dateTimeProvider = (SystemDateTimeProviderStub)GetService<ISystemDateTimeProvider>();
     }
 
+    /// <summary>
+    /// This is implemented to support the "hack" where
+    ///     MeteredDataResponsible is working as GridOperator.
+    /// </summary>
+    [Fact]
+    public async Task
+        Given_OutgoingEnergyResultMessageToMeteredDataResponsible_When_DelegatedByIsGridOperator_Then_GridOperatorReceivesMessage()
```
MWO1024, line 64: Assuming 'given' is for prerequisites, 'when' is the action, and 'then' is the result, maybe "Given_OutgoingMeteredDataResponsibleMessageAndDelegationBetweenGridOperatorRoles_When_EnqueuingMeteredDataResponsibleMessage_Then_MessageEnqueuedToDelegatedGridOperator"?

source/MasterData.Application/MasterDataClient.cs
```diff
@@ -180,6 +181,18 @@ await _actorCertificateRepository
             processType,
             cancellationToken).ConfigureAwait(false);
 
+        if (processDelegation is null
+            && WorkaroundFlags.MeteredDataResponsibleToGridOperatorHack
+            && delegatedByActorRole.Equals(ActorRole.MeteredDataResponsible))
+        {
+            processDelegation = await _processDelegationRepository.GetAsync(
+                delegatedByActorNumber,
+                delegatedByActorRole.ForActorMessageQueue(),
```
RasmusGraabaek, line 190:
```suggestion
                delegatedByActorRole.TransformRoleIfMdrHack(),
```

RasmusGraabaek (approved):

PR #964, opened 2024-04-15
feat: Add EventIds to ArchivedMessages table

ebbeknudsen:
## Description

Add `EventIds` column to archived messages, which describes the `eventId`'s of the outgoing messages that was used to create the archived message (there can be more than one, if we're bundling messages). This column is purely for debugging, so instead of creating a one-to-many relation table, we've just used a `::` seperated column.

The `EventId` describes the Service Bus MessageId property, so it can be searched for in application insights, to debug the application flow regarding the message.


source/IntegrationTests/Application/IncomingMessages/WhenIncomingMessagesIsReceivedTests.cs
```diff
@@ -325,10 +331,39 @@ public async Task Incoming_message_is_archived_with_correct_data(string path, st
             CancellationToken.None);
 
         // Assert
-        var archivedMessageId = await GetArchivedMessageIdFromDatabaseAsync(messageIdFromFile);
-        var archivedMessageFileStorageReference = await GetArchivedMessageFileStorageReferenceFromDatabaseAsync(messageIdFromFile);
-        archivedMessageFileStorageReference.Should().Be($"{senderActorNumber.Value}/{year:0000}/{month:00}/{date:00}/{archivedMessageId:N}");
-        await AssertArchivedMessageDocumentType(messageIdFromFile, incomingDocumentType);
+        // var archivedMessageId = await GetArchivedMessageIdFromDatabaseAsync(messageIdFromFile);
```
MadsDue, line 334: remove comments

source/IntegrationTests/Application/IncomingMessages/WhenIncomingMessagesIsReceivedTests.cs
```diff
@@ -325,10 +331,39 @@ public async Task Incoming_message_is_archived_with_correct_data(string path, st
...
+        // var archivedMessageFileStorageReference = await GetArchivedMessageFileStorageReferenceFromDatabaseAsync(messageIdFromFile);
+        // await AssertArchivedMessageDocumentType(messageIdFromFile, incomingDocumentType);
```
MadsDue, line 336: Are AssertArchivedMessageDocumentType not needed anymore and can be deleted, or should it be updated with below and be used for both tests?

ebbeknudsen: Correct, it is no longer needed, since it is asserted when asserting all archived message properties. I've deleted the method 👍

MadsDue (approved):

PR #970, opened 2024-04-18
bug: ensure outgoing message file storage reference is updated to match delegation

MadsDue:
## Description
We should ensure that the file storage reference for outgoing messages is matching the receiver.


source/OutgoingMessages.Infrastructure/OutgoingMessages/OutgoingMessageEntityConfiguration.cs
```diff
@@ -116,7 +116,7 @@ public void Configure(EntityTypeBuilder<OutgoingMessage> builder)
                 });
 
             builder.Property<string>("CreatedBy");
-            builder.Property<Instant>("CreatedAt");
+            builder.Property(x => x.CreatedAt);
```
RasmusGraabaek, line 119: Would it make sence to move "createdAt" up to the other properties. Since it's not a "shadow"-field anymore?

MWO1024 (dismissed):

RasmusGraabaek (approved):

PR #985, opened 2024-04-25
chores: added retry comments on service bus listners

MadsDue:


source/B2BApi/EventListeners/IntegrationEventListener.cs
```diff
@@ -34,6 +34,14 @@ public IntegrationEventListener(ILogger<IntegrationEventListener> logger, ISubsc
         _subscriber = subscriber;
     }
 
+    /// <summary>
+    /// Receives messages from the inbox queue and forwards them to the inbox event receiver.
```
RasmusGraabaek, line 38: no inbox here?

PR #992, opened 2024-04-30
test: Add behaviour test for delegation synchronous error

ebbeknudsen:
## Description

- Refactor `IncomingMessageClient.RegisterAndSendAsync()` `responseFormat` to make it more intuitive
- Add behaviour tests for synchronous validation error when delegation can't find the original actor number (for example if the role is energy supplier and energy supplier is not supplied in the message)

### DIFF
--- a/source/IncomingMessages.Application/IncomingMessageClient.cs
+++ b/source/IncomingMessages.Application/IncomingMessageClient.cs
@@ -69,8 +69,8 @@ public class IncomingMessageClient : IIncomingMessageClient
         IIncomingMessageStream incomingMessageStream,
         DocumentFormat documentFormat,
         IncomingDocumentType documentType,
-        CancellationToken cancellationToken,
-        DocumentFormat responseFormat = null!)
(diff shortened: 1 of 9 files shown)


MWO1024 (approved):

source/IncomingMessages.Application/IncomingMessageClient.cs
```diff
@@ -67,16 +67,16 @@ public class IncomingMessageClient : IIncomingMessageClient
 
     public async Task<ResponseMessage> RegisterAndSendAsync(
         IIncomingMessageStream incomingMessageStream,
-        DocumentFormat documentFormat,
+        DocumentFormat incomingDocumentFormat,
         IncomingDocumentType documentType,
-        CancellationToken cancellationToken,
-        DocumentFormat responseFormat = null!)
+        DocumentFormat responseDocumentFormat,
```
MadsDue, line 72: First time I read it, I was a little confused about our incoming message client should return in a DocumentFormat, since it is not returning documents :/

PR #1003, opened 2024-05-02
chore: Remove deprecated energinet.core nuget packages

RasmusGraabaek:


source/B2CWebApi/Program.cs
```diff
@@ -41,8 +40,6 @@
 
     // Logging
     .AddApplicationInsightsForWebApp(domainName)
```
dstenroejl, line 42: Let us rename "domainName" to "subsystemName" :)

PR #1006, opened 2024-05-03
test: Add behaviour test for requests with multiple series

ebbeknudsen:

### DIFF
--- a/source/IntegrationTests/Behaviours/WholesaleServicesBehaviourTestBase.cs
+++ b/source/IntegrationTests/Behaviours/WholesaleServicesBehaviourTestBase.cs
@@ -100,80 +100,44 @@ public abstract class WholesaleServicesBehaviourTestBase : BehavioursTestBase
         return GivenWholesaleServicesRequestResponseIsReceived(processId, rejectedMessage);
     }
 
-    protected Task GivenWholesaleServicesRequestResponseIsReceived<TType>(Guid processId, TType wholesaleServicesRequestResponseMessage)
-        where TType : IMessage
+    protected async Task<(WholesaleServicesRequest WholesaleServicesRequest, Guid ProcessId)> ThenWholesaleServicesRequestServiceBusMessageIsCorrect(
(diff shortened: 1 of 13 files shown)


source/IntegrationTests/TestDoubles/ServiceBusSenderSpy.cs
```diff
@@ -34,14 +36,19 @@ public ServiceBusSenderSpy(string topicName)
...
+
         public bool MessageSent { get; private set; }
 
         public Task SendAsync(ServiceBusMessage message, CancellationToken cancellationToken)
         {
             if (ShouldFail)
                 throw new ServiceBusException();
+
+            MessagesSent.Add(message);
             Message = message;
```
RasmusGraabaek, line 49: Rename?

source/IntegrationTests/Behaviours/WholesaleServicesBehaviourTestBase.cs
```diff
@@ -253,4 +225,64 @@ protected async Task WhenWholesaleServicesProcessIsInitialized(ServiceBusMessage
...
+                    .BeEquivalentTo(
+                        input.ChargeTypes.Select(
+                            ct => new Energinet.DataHub.Edi.Requests.ChargeType
+                            {
+                                ChargeType_ = ct.ChargeType, ChargeCode = ct.ChargeCode,
+                            }));
+            }
+
+            message.PeriodStart.Should().Be(input.Period.Start.ToString());
+            message.PeriodEnd.Should().Be(input.Period.End.ToString());
```
RasmusGraabaek, line 271: End is nullable?

ebbeknudsen: It is allowed as nullable but required in validation right? Which is why i didn't make it nullable in these tests, since it's not a valid request

RasmusGraabaek (approved):

PR #1016, opened 2024-05-15
test: Wait for integration events in acceptance test

ebbeknudsen:
## Description

We sometimes experience tests failing because we create an integration event, but do not wait for it to be handled. This happens especially around ebIX and `ActorCertificateCredentialsAssigned`/`ActorCertificateCredentialsRemoved`. Since our acceptance tests now access the database to create processes, we can also implement waiting for an integration event to be received.

- Fix disposing acceptance tests
- Wait for integration event to be handled in acceptance tests


source/AcceptanceTests/Drivers/IntegrationEventPublisher.cs
```diff
@@ -39,9 +43,41 @@ public async ValueTask DisposeAsync()
...
     }
 
-    internal Task PublishAsync(string eventName, byte[] eventPayload)
+    internal async Task PublishAsync(string eventName, byte[] eventPayload, bool waitForHandled)
     {
-        return _sender.SendMessageAsync(CreateIntegrationEventMessage(eventName, eventPayload));
+        var integrationEvent = CreateIntegrationEventMessage(eventName, eventPayload);
+        await _sender
+            .SendMessageAsync(integrationEvent)
+            .ConfigureAwait(false);
```
MWO1024, line 51: should we use `.ConfigureAwait` in tests? I don't think we usually do.

MWO1024 (approved):

source/AcceptanceTests/Drivers/IntegrationEventPublisher.cs
MWO1024: Fair enough, it seems like we do in acceptance tests

ebbeknudsen: Right now we use it in the classes that are not the `XyzTest` classes themselves 👍

PR #1017, opened 2024-05-15
feat: Add B2C endpoint for wholesale settlement requests

MWO1024:


source/B2CWebApi/Factories/RequestWholesaleSettlementDtoFactory.cs
```diff
@@ -0,0 +1,123 @@
...
+using Energinet.DataHub.EDI.BuildingBlocks.Domain.Models;
+using Energinet.DataHub.EDI.IncomingMessages.Interfaces;
+using NodaTime;
+
+namespace Energinet.DataHub.EDI.B2CWebApi.Factories;
+
+public static class RequestWholesaleSettlementDtoFactory
+{
+    // TODO (MWO): fix value
+    private const string WholesaleSettlementMessageType = "E74";
```
ebbeknudsen, line 26: https://energinet.sharepoint.com/sites/DH3ART-team/Delte%20dokumenter/Forms/AllItems.aspx?id=%2Fsites%2FDH3ART%2Dteam%2FDelte%20dokumenter%2FGeneral%2FCIM%2FCIM%20XSD%20%2D%20XML%2FXML%20filer%2FXML%2020220706%20%2D%20Danske%20koder%20%2D%20v%2E1%2E5

**D21**

source/B2CWebApi/Factories/RequestWholesaleSettlementDtoFactory.cs
```diff
@@ -0,0 +1,122 @@
...
+        return new RequestWholesaleSettlementDto(
+            senderNumber,
+            senderRoleCode,
+            DataHubDetails.DataHubActorNumber.Value,
+            MarketRole.CalculationResponsibleRole.Code,
+            MapToBusinessReasonCode(request.CalculationType),
+            WholesaleSettlementMessageType,
+            Guid.NewGuid().ToString(),
+            now.ToString(),
+            Electricity,
```
lasrinnil, line 60: Are we sure that the `BusinessType` for WholesaleSettlement is `Electricity`?

source/B2CWebApi/Models/RequestAggregatedMeasureDataMarketDocument.cs
```diff
@@ -23,21 +23,11 @@ public enum MeteringPointType
     Exchange,
```
lasrinnil, line 23: Should we eliminate the `MeteringPointType` here aswell?

lasrinnil (approved):

source/B2CWebApi/Models/RequestAggregatedMeasureDataMarketDocument.cs
MWO1024: We could. I moved CalculationType because thet both needed it. WholesaleSettlement doesnt require metering point type, so I left it as-is.

source/B2CWebApi/Factories/RequestWholesaleSettlementDtoFactory.cs
MWO1024: Good question. I'll look in an example file.

MWO1024: According to an example from Ebbes link above, it is.

source/B2CWebApi/Factories/RequestWholesaleSettlementDtoFactory.cs
```diff
@@ -0,0 +1,122 @@
...
+        string senderRole,
+        DateTimeZone dateTimeZone,
+        Instant now)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+
+        var senderRoleCode = MapRoleNameToCode(senderRole);
+
+        var series = new RequestWholesaleSettlementSeries(
+            Guid.NewGuid().ToString(),
```
MadsDue, line 41:
```suggestion
            TransactionId.New().Value,
```

source/B2CWebApi/Factories/RequestWholesaleSettlementDtoFactory.cs
MadsDue: Same in the example we use in our acceptance tests, which we got from the business

source/B2CWebApi/Factories/RequestWholesaleSettlementDtoFactory.cs
```diff
@@ -0,0 +1,122 @@
...
+            now.ToString(),
+            Electricity,
+            new[] { series });
+    }
+
+    private static string? SetSettlementVersion(CalculationType calculationType)
+    {
+        if (calculationType == CalculationType.FirstCorrection)
+        {
+            return "D01";
```
MadsDue, line 68: Consider, doing this for all corrections

```suggestion
            return SettlementVersion.FirstCorrection.Code;
```

Like you do for BusinessReason below.

source/B2CWebApi/Models/RequestWholesaleServicesMarketDocument.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Energinet.DataHub.EDI.B2CWebApi.Models;
+
+/// <summary>
+///     Responsible for carrying the market message data from the incoming message before any data validation.
+/// </summary>
+public sealed record RequestWholesaleSettlementMarketRequest(
+    CalculationType CalculationType,
+    string StartDate,
+    string EndDate,
```
MadsDue, line 23: Is EndDate not nullable or can't not be null from the UI?
```suggestion
    string? EndDate,
```
It looks like RequestWholesaleSettlementDtoFactory is treating it as nullable

source/Tests/B2CWebApi/RequestAggregatedMeasureData/RequestAggregatedMeasureDataFactoryTests.cs
MadsDue, line 1: Should we also have those tests case for wholesaleSettlementFactory?

MWO1024: I have added a test. Not covering eveything, as the risk is set to low.

source/B2CWebApi/Models/RequestWholesaleServicesMarketDocument.cs
MWO1024: The same applies to aggregations iirc. I made them "identical" to make things easier to understand. I'm tempted to keep it as-is to avoid having to rerun the things in BFF and to keep it consistent. Ideally, I would prefer if we changed it in such a way that we skip the internal DTO and just translate to the B2B one.

MadsDue (approved):

PR #1027, opened 2024-05-23
docs: Documentation of test naming convention

RasmusGraabaek:


source/IntegrationTests/README.md
```diff
@@ -0,0 +1,123 @@
+# EDI integration test project
+
+As of right now, the following naming convention is [automatically enforced](./Behaviours/MetaTests.cs) for tests
+located in the `Behaviours` folder.
+But not for tests located outside of this folder, it is therefore up to the individual to enfore the descriped
+guidelines/convention
```
MWO1024, line 6: It is not enforced for tests located outside of this particular folder, and it is thus the responsibility of the developer to enforce and adhere to the following guidelines.

source/IntegrationTests/README.md
```diff
@@ -0,0 +1,123 @@
...
+}
+```
+
+That is. Most, if not all, of the variable are defined as the first thing.
+
+Hereafter executing the method(s) one wants to test. Save the return value in a variable with a descriptive name,
+if this is not doable use `actual` as the variable name.
+
+Then the assertions begin.
+When asserting it is preferred to use `FluentAssertion` and make use of `AssertionScope()`, as in the example above.
```
MWO1024, line 33: That is, we are adhering to the well-established "Arrange, Act, and Assert"-pattern (AAA, 3As, ...). There exists a plethora of information and guidelines regarding AAA, but for the uninitiated, it can be summarised---in the words of Vladimir Khorikov---as such:
<quote>
- In the _arrange section_, you bring the system under test (SUT) and its dependencies to a desired state
- In the _act section_, you call methods on the SUT, pass the prepared dependencies, and capture the output value (if any)
- In the _assert section_, you verify the outcome. The outcome may be represented by a return value, the final state of the SUT and its collaborators, or the methods the SUT called on those collaborators
<end quote>

We have a few EDI-specific conventions, in addition to the established AAA conventions:
- If no suitable name can be given to the return value, `actual` or `result` are good fallback values
- When doing assertions, we much prefer the usage of `FluentAssertion` and the utilisation of `AssertionScope`. This is to ensure the assertions are readable, gives good error messages, and that a given test return all errors immediately.

source/IntegrationTests/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Naming the tests
+
+The structure described in the [section above](#tests-structure) give rise to the following naming
+schema `Given_XXX_When_YYY_Then_ZZZ`.
+Where:
+
+`XXX` states the important variables (Related to the `//Arrange` section of the test).
```
MWO1024, line 41: "important set-up"? "important configuration"? "the business state represented by the arrange section"?

source/IntegrationTests/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+`YYY` is the action or a description of the action that is tested (Related to the `//Act` section of the test).
+
+`ZZZ` is a description of the expected consequence of the action stated in `YYY` (Related to the `//Assert` section of the test).
+
+### Alternative test naming
+
+If one from whatever reason should follow another naming convention than stated in the [naming section](#naming-the-tests) one should add the `[ExcludeFromNameConventionCheck]` attribute to the test. As is done in the following example:
```
MWO1024, line 49: If, for whatever (good) reason, one desire to deviate from the [naming convention](#naming-the-tests), one should utilise the `[ExcludeFromNameConventionCheck]` attribute. This should however not be done lightly, and adding the justification in a comment to the test is recommended.

source/IntegrationTests/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+```cs
+    [ExcludeFromNameConventionCheck]
```
MWO1024, line 52: We should then add
`// Some sort of justification in the form of a comment`
or similar to the example

source/IntegrationTests/README.md
```diff
@@ -0,0 +1,123 @@
...
+    public static void ThisDoesNotSatisfyTheNamingConventionFact()
+    {
+        
+    }
+```
+
+Which serves two purposes.
+
+- Firstly: That the test is excluded for the [naming convention check](./Behaviours/MetaTests.cs).
+- Secondly: It makes it clear that one purposely did not follow the naming convention. And hopefully has a good reason for doing so.
```
MWO1024, line 62: Might be redundant assuming something like the above is added? If not, I'll look at this again.

source/IntegrationTests/README.md
RasmusGraabaek: Using "important set-up"

source/IntegrationTests/README.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+## Folder structure
+
+### Behaviours test
+
+The `Behaviours` test have a special kind of folder structure.
+Since these kind of tests are supposed to test the behavior of the system given some external input.
+Which gives rise to a different kind of class naming, namely that the classes are named as the following: `Given{ExternalInput}Tests`.
+eg:
```
MWO1024, line 77: Our `Behaviours` tests are additionally subject to further restrictions (or guidelines if you prefer) wrt folder structure and file naming. As the primary goal of these tests are to validate the behaviour of the (sub-)system given some external stimulus (e.g. an integration event or request), this external stimulus has to be part of the name of the given test class. All tests for the same external stimulus are therefore located in the same test file. In other words: all behaviour tests must be named as follows: `Given{ExternalInput}Tests`. 

Furthermore, the tests must be group according to the kind of stimulus we are testing: integration events, incoming requests, etc.. This is achieved by folders `IntegrationEvents`, `IncomingRequests`, etc., directly in the root of `Behaviours`. The combination of these two rules are illustrated below with an example.

source/IntegrationTests/README.md
```diff
@@ -0,0 +1,129 @@
...
+|   |   ├── GivenMonthlyAmountPerChargeResultProducedV1ReceivedTests.cs
+│   └── IncomingRequests
+│       ├── GivenAggregatedMeasureDataRequestTests.cs
+|       └── GivenAggregatedMeasureDataRequestWithDelegationTests.cs
+├── DocumentAsserters
+├── EventBuilders
+```
+
+By doing so, one does not need to specify the external input in the specific test method. Which allow the writer to
+specify the important variables in the test, as descripbed in [naming section](#naming-the-tests).
```
MWO1024, line 92: By doing so, one does not have to "waste" space on specifying the external input in each and every single test method's [`Given`-part](#naming-the-tests), but can instead focus on the uniqueness of each given test instead.

source/IntegrationTests/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+If one finds it unpleasent to have a class named `Given{ExternalInput}Tests` with a method named `Given{ImportantVairalbes}_When_YYY_Then_ZZZ`.
+Then it is possible the use the prefix `AndGiven_` instead of `Given_`. It is even possible to leave it out and start
+with `When_`.
```
MWO1024, line 96: Similarily, if one finds it unpleasent to have a class named `Given{ExternalInput}Tests` with a method named `Given{ImportantSetup}_When_YYY_Then_ZZZ`, it is possible the use the prefix `AndGiven_` instead of `Given_`. In the cases where there are no additional setup, it is possible to start with `When_` directly. However, we advise caution in this case, as it is rare for a test to have no setup.

source/IntegrationTests/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Propper integration tests
```
MWO1024, line 98: Proper

source/IntegrationTests/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+Unlike the `Behaiviours` tests these tests are more isolated and test a single responsibility. Eg test of
+integrationEventHandlers, repositories and the like.
+
+Since the responsibility is different, so is the folder structure.
+In fact, there isn't any good structure for the time being :I.
+
+But one should try to keep the test for a single module to it's own folder.
+eg:
```
MWO1024, line 107: That said, one should try to keep the test for a single module to it's own folder. This way they can be extracted to their own modules eventually, with relative ease.

source/IntegrationTests/README.md
```diff
@@ -0,0 +1,129 @@
...
+├── Infrastructure
+├── DocumentAsserters
+├── EventBuilders
+```
+
+Here `Trancastions` are supposed to test the `process` module and the `integrationEvent` module. Which should be moved
+to a more specialized test project at some point
+
+Note that the naming convention mentioned in the [naming section](#naming-the-tests) is not automatically enforced but
+it is prefer that one tries to follow it anyway.
```
MWO1024, line 129: In the example above, the tests in the `Transactions` folder are testing both the `Process` and `IntegrationEvent` module. Ideally, if possible, the tests that are limited to a single module should be moved to that modules respective folder (or actual test module). For the cross-module integration tests, no well-established guideline currently exists.

Do note, that while the [naming convention](#naming-the-tests) is not enforced here, it is advisable to adhere to it as much as possible. This ensures a consistent naming across the solution.

lasrinnil (approved):
If the inspiration for the GWT statements are from Gherkin. Perhaps we should mention it for context.

PR #1038, opened 2024-05-31
fix: Allow nullable charge type ids in requests

MWO1024:


RasmusGraabaek (dismissed):

source/Process.Infrastructure/Transactions/WholesaleServices/WholesaleServicesRequestFactory.cs
```diff
@@ -72,16 +70,20 @@ private static WholesaleServicesRequest CreateWholesaleServicesRequest(Wholesale
...
         {
             request.ChargeTypes.Add(
-                new ChargeType() { ChargeCode = chargeType.Id, ChargeType_ = MapChargeType(chargeType.Type) });
+                (chargeType.Id, chargeType.Type) switch
+                {
+                    (null, null) => new ChargeType(),
+                    (null, var type) => new ChargeType { ChargeType_ = MapChargeType(type) },
+                    (var id, null) => new ChargeType { ChargeCode = id },
+                    var (id, type) => new ChargeType { ChargeCode = id, ChargeType_ = MapChargeType(type) },
+                });
```
ebbeknudsen, line 79: I think it would be more readable (and follow the existing pattern) to do something like this:
```csharp
var mappedChargeType = new ChargeType();
if (chargeType.Id is not null)
    mappedChargeType.ChargeCode = Id;

if (chargeType.Type is not null)
    mappedChargeType.ChargeType_ = MapChargeType(type)
```

source/Tests/Infrastructure/IncomingMessages/WholesaleServicesRequestFactoryTests.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        // Assert
+        serviceBusMessage.Should().NotBeNull();
+        serviceBusMessage.Body.Should().NotBeNull();
+        serviceBusMessage.Subject.Should().Be("WholesaleServicesRequest");
+
+        WholesaleServicesRequest.Parser.ParseFrom(serviceBusMessage.Body)
+            .ChargeTypes
+            .Should()
+            .ContainSingle();
```
ebbeknudsen, line 69: Should we assert the code and type is set to the correct value on the charge type?

MWO1024: I have added some validation. We should probably extend the tests to cover the entire factory at some point though.

lasrinnil (approved):

PR #1039, opened 2024-05-31
refac: ebix messageheader id has no dashes

lasrinnil:


lasrinnil:
Approch discarded by the team

PR #1059, opened 2024-06-10
refac: Peek and Bundle clean-up

MWO1024:


source/IntegrationTests/Application/OutgoingMessages/WhenAPeekIsRequestedTests.cs
```diff
@@ -300,9 +300,9 @@ public async Task Given_OutgoingMessage_When_MessageIsPeeked_Then_MessageIsArchi
 
         // Assert / Then
         peekResult.Should().NotBeNull("because a peek result should be found");
-        peekResult.MessageId.Should().NotBeNull("because a peek result with a message id should be found");
+        peekResult!.MessageId.Should().NotBeNull("because a peek result with a message id should be found");
```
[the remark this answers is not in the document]
MWO1024: It only knows after the first access. Since `.Should()` handles null values, it isn't guaranteed to be not null at this point :/ But any following access to `peekResult` should not be null (otherwise we would have received a NPE at `!`).

source/Tests/Domain/OutgoingMessages/Queueing/ActorMessageQueueTests.cs
```diff
@@ -76,10 +83,14 @@ public void Peek_returns_empty_bundle_if_bundle_has_been_dequeued()
         var outgoingMessage = CreateOutgoingMessage(receiver, BusinessReason.BalanceFixing);
         actorMessageQueue.Enqueue(outgoingMessage, SystemClock.Instance.GetCurrentInstant());
 
-        var result = actorMessageQueue.Peek();
+        var peekResult = actorMessageQueue.Peek();
+        peekResult.Should().NotBeNull();
+
+        var dequeueResult = actorMessageQueue.Dequeue(peekResult!.MessageId);
+        dequeueResult.Should().BeTrue("we should be able to dequeue what we just peeked");
```
RasmusGraabaek, line 90:
```suggestion
        dequeueResult.Should().BeTrue("We should be able to dequeue what we just peeked");
```

source/Tests/Domain/OutgoingMessages/Queueing/ActorMessageQueueTests.cs
```diff
@@ -91,12 +102,15 @@ public void If_current_bundle_is_full_the_message_is_assigned_to_a_new_bundle()
         actorMessageQueue.Enqueue(CreateOutgoingMessage(receiver, BusinessReason.BalanceFixing), SystemClock.Instance.GetCurrentInstant(), maxNumberOfMessagesInABundle: 1);
 
         var firstBundle = actorMessageQueue.Peek();
-        actorMessageQueue.Dequeue(firstBundle.MessageId!.Value);
+        firstBundle.Should().NotBeNull();
+
+        var dequeueResult = actorMessageQueue.Dequeue(firstBundle!.MessageId);
+        dequeueResult.Should().BeTrue("we should be able to dequeue what we just peeked");
```
RasmusGraabaek, line 108:
```suggestion
        dequeueResult.Should().BeTrue("We should be able to dequeue what we just peeked");
```

source/Tests/Domain/OutgoingMessages/Queueing/ActorMessageQueueTests.cs
```diff
@@ -107,14 +121,16 @@ public void Messages_are_bundled_by_message_type_and_process_type()
...
 
-        var firstPeekResult = actorMessageQueue.Peek(MessageCategory.Aggregations);
-        actorMessageQueue.Dequeue(firstPeekResult.MessageId!.Value);
-        var secondPeekResult = actorMessageQueue.Peek(MessageCategory.Aggregations);
-        actorMessageQueue.Dequeue(secondPeekResult.MessageId!.Value);
+        var firstBundle = actorMessageQueue.Peek();
+        firstBundle.Should().NotBeNull();
+
+        var dequeueResult = actorMessageQueue.Dequeue(firstBundle!.MessageId);
+        dequeueResult.Should().BeTrue("we should be able to dequeue what we just peeked");
```
RasmusGraabaek, line 128:
```suggestion
        dequeueResult.Should().BeTrue("We should be able to dequeue what we just peeked");
```

RasmusGraabaek (dismissed):

source/OutgoingMessages.Domain/Models/PeekResult.cs
```diff
@@ -17,4 +17,4 @@
 
 namespace Energinet.DataHub.EDI.OutgoingMessages.Domain.Models;
 
-public sealed record PeekResult(BundleId? BundleId, MessageId? MessageId, DocumentType? DocumentType);
+public sealed record PeekResult(BundleId BundleId, MessageId MessageId);
```
MadsDue, line 20: You could consider adding a description on this :)

MadsDue (dismissed):

source/Tests/Domain/OutgoingMessages/Queueing/ActorMessageQueueTests.cs
MWO1024: We shouldn't use upper case for these errors, as they are going to be present mid-sentence right after a 'because':

> Expected dequeueResult to be false because we should be able to dequeue what we just peeked, but found True.

source/OutgoingMessages.Domain/Models/PeekResult.cs
MWO1024: I have added a description. Hopefully it is sensible.

RasmusGraabaek (approved):

PR #1068, opened 2024-06-12
feat: Add additional activities to query energy results

dstenroejl:
## Description

- [x] Add activities for all energy result queries
- [x] Use fan-out/fan-in for activities
- [x] Update orchestration test

## References

Part of:
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/199


source/B2BApi.AppTests/Functions/EnqueueMessages/EnqueueMessagesOrchestrationTests.cs
```diff
@@ -135,10 +137,12 @@ public async Task Given_FeatureFlagIsEnabledAndCalculationOrchestrationId_When_C
             .OrderBy(item => item["Timestamp"])
             .Select(item => item.Value<string>("FunctionName"));
 
-        activities.Should().NotBeNull().And.Equal(
+        activities.Should().NotBeNull().And.BeEquivalentTo(
```
RasmusGraabaek, line 140: Should we have an assertions of that the second to last activity is "SendActorMessagesEnqueuedActivity" ?

RasmusGraabaek (approved):

source/B2BApi.AppTests/Functions/EnqueueMessages/EnqueueMessagesOrchestrationTests.cs
ebbeknudsen: We have a TODO on the #5.1 task to add more assertions in this test later. We can also test that the activity results has the expected enqueued actor messages count - but we will do that in a later PR

ebbeknudsen (approved):

PR #1078, opened 2024-06-20
refac: Outgoingmessage idempotency

lasrinnil:
## Description
This PR seeks to ensure that messages are only enqueued once pr: actor number + actor role + external id.
- Added `ExternalId` to `dbo.OutgoingMessages`
- Added Unique index on `dbo.OutgoingMessages` on (ExternalId, ReceiverNumber, ReceiverRole)
- Added `ExternalId` to `OutgoingMessage`
- Updated `EnqueueMessage` usecase to check if an `OutgoingMessage` excist for the `Receiver` and `ExternalId` when called to ensure that we only enqueue a message once then.


source/OutgoingMessages.Interfaces/Models/WholesaleServicesTotalSumMessageDto.cs
```diff
@@ -42,6 +42,7 @@ public class WholesaleServicesTotalSumMessageDto : OutgoingMessageDto
             receiverRole,
             DataHubDetails.DataHubActorNumber,
             ActorRole.MeteredDataAdministrator,
+            new ExternalId(Guid.NewGuid()), // TODO: LRN Should come from calculationresultid
```
lasrinnil, line 45: This will be replaced before merging

source/BuildingBlocks.Domain/Models/ExternalId.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Energinet.DataHub.EDI.BuildingBlocks.Domain.Models;
+
+/// <summary>
+/// An ExternalId to represent what the systems external source uses as an identifier
+/// </summary>
+[Serializable]
+public record ExternalId
+{
+    [JsonConstructor]
+    public ExternalId(Guid value)
```
RasmusGraabaek, line 26:
```suggestion
    public ExternalId(Guid value)
```
Since we have a "From" method, do we want a private constructure?

source/IntegrationTests/Application/OutgoingMessages/WhenEnqueueingOutgoingMessageTests.cs
```diff
@@ -360,6 +361,75 @@ public async Task Given_EnqueuingNotifyAggregatedMeasureData_When_ReceiverActorR
         fromDb.OutgoingMessageReceiverRole.Should().Be(ActorRole.MeteredDataResponsible.Code);
     }
 
+    [Fact]
+    public async Task Outgoing_message_must_only_be_enqueued_once_to_an_receiver_with_externalId()
```
RasmusGraabaek, line 365: The naming does not adhere to the naming convention: https://github.com/Energinet-DataHub/opengeh-edi/blob/main/source/IntegrationTests/README.md

RasmusGraabaek: But it is not enforced here :)

lasrinnil: then it will not be fixed :)

RasmusGraabaek (dismissed):

RasmusGraabaek (approved):

source/IntegrationTests/Application/OutgoingMessages/WhenEnqueueingOutgoingMessageTests.cs
```diff
@@ -360,6 +365,61 @@ public async Task Given_EnqueuingNotifyAggregatedMeasureData_When_ReceiverActorR
...
+    [Fact]
+    public async Task Outgoing_message_must_only_be_enqueued_once_to_an_receiver_with_same_externalId()
+    {
+        // Arrange
+        var receiverId = ActorNumber.Create("1234567891912");
+        var externalId = Guid.NewGuid();
+
+        var message = _wholesaleAmountPerChargeDtoBuilder
+            .WithReceiverNumber(receiverId.Value)
+            .WithCalculationResultId(externalId)
```
MadsDue, line 377: Should we extend out dto with an external id field instead? 
CalculationResult id is not equal to Outgoing_message_must_only_be_enqueued_once_to_an_receiver_with_same_externalId in all cases

E.g. if there is a wholesale periode with lacking fees, then we need to split the result from a single calculation result id into multiple messages

source/OutgoingMessages.Domain/Models/OutgoingMessages/IOutgoingMessageRepository.cs
```diff
@@ -32,4 +31,9 @@ public interface IOutgoingMessageRepository
     /// Get all messages assigned to a bundle by id
     /// </summary>
     Task<OutgoingMessageBundle> GetAsync(PeekResult peekResult);
+
+    /// <summary>
+    /// Checks if a message exists in the database for the designated receiver and the external id.
+    /// </summary>
+    Task<OutgoingMessage?> GetAsync(Receiver receiver, ExternalId externalId);
```
MadsDue, line 38: Summary dosn't match method name

source/IntegrationTests/Application/OutgoingMessages/WhenEnqueueingOutgoingMessageTests.cs
```diff
@@ -360,6 +365,61 @@ public async Task Given_EnqueuingNotifyAggregatedMeasureData_When_ReceiverActorR
         fromDb.OutgoingMessageReceiverRole.Should().Be(ActorRole.MeteredDataResponsible.Code);
     }
 
+    [Fact]
+    public async Task Outgoing_message_must_only_be_enqueued_once_to_an_receiver_with_same_externalId()
```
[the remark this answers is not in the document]
lasrinnil: done

source/OutgoingMessages.Application/UseCases/EnqueueMessage.cs
```diff
@@ -64,6 +64,10 @@ public class EnqueueMessage
                 .ConfigureAwait(false);
         }
 
+        var messageExists = await _outgoingMessageRepository.GetAsync(messageToEnqueue.Receiver, messageToEnqueue.ExternalId).ConfigureAwait(false);
```
MadsDue, line 67: Is this enough? 

Shouldn't we have an index on this as well, if you message with same External id is enqueued at the same time with two different db contexts?

source/OutgoingMessages.Domain/Models/OutgoingMessages/IOutgoingMessageRepository.cs
lasrinnil: ✔️

source/OutgoingMessages.Application/UseCases/EnqueueMessage.cs
lasrinnil: CREATE UNIQUE INDEX IDX_OutgoingMessageIdempotency ON [dbo].[OutgoingMessages] (ExternalId, ReceiverNumber, ReceiverRole);

MadsDue (approved):

PR #1088, opened 2024-06-26
refac: Bundle PeekedAt

lasrinnil:
## Description
To be more explicit of when a `Bundle` actually have been peeked by an actor - this PR will add a `PeekedAt` property to `Bundle` which will be updated with a `Instant` of when the Actor peeked the `Bundle`

### DIFF
--- a/source/OutgoingMessages.Domain/Models/ActorMessagesQueues/ActorMessageQueue.cs
+++ b/source/OutgoingMessages.Domain/Models/ActorMessagesQueues/ActorMessageQueue.cs
@@ -105,7 +105,6 @@ public class ActorMessageQueue
     {
         return _bundles.FirstOrDefault(bundle =>
             bundle.ClosedAt is null
-            && bundle.PeekedAt is null
             && bundle.DocumentTypeInBundle == messageType
             && bundle.BusinessReason == businessReason
             && bundle.RelatedToMessageId?.Value == relatedToMessageId?.Value);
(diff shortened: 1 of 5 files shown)


source/OutgoingMessages.Domain/Models/ActorMessagesQueues/ActorMessageQueue.cs
```diff
@@ -105,6 +105,7 @@ private void EnsureApplicable(OutgoingMessage outgoingMessage)
     {
         return _bundles.FirstOrDefault(bundle =>
             bundle.ClosedAt is null
+            && bundle.PeekedAt is null
```
MadsDue, line 108: Should this not be aligned with the bundle.Add() method. Currently, the Add only skip if the bundle is closed. 
Accordingly to this change the bundle.PeekedAt can be not null and ClosedAt null. This will return the current bundle (even if it is closed) and not allow the outgoing message to be added to the bundle and the outgoing message will never be enqueued

source/OutgoingMessages.Domain/Models/Bundles/Bundle.cs
```diff
@@ -66,9 +68,10 @@ private Bundle()
 
     internal Instant? ClosedAt { get; private set; }
 
-    public void CloseBundle()
+    public void PeekBundle()
     {
         ClosedAt = SystemClock.Instance.GetCurrentInstant();
+        PeekedAt = SystemClock.Instance.GetCurrentInstant();
```
MadsDue, line 74: Should we update the closedAt when PeekBundle is triggered?

lasrinnil: Only if it is not set. fixed

MadsDue (approved):
LGTM

PR #1145, opened 2024-08-07
feat: handle response from wholesale monthly request with total monthly amount

MadsDue:
## Description
Ensure EDI is prepared to handle wholesale requests for the total monthly amount.


source/Process.Application/Transactions/WholesaleServices/WholesaleServicesRequestAcceptedMapper.cs
```diff
@@ -47,18 +47,19 @@ public Task<INotification> MapFromAsync(byte[] payload, EventId eventId, Guid re
...
                 SettlementVersion: MapSettlementVersion(wholesaleSeries.CalculationType),
                 MapSettlementMethod(wholesaleSeries.SettlementMethod),
                 MapCurrency(wholesaleSeries.Currency),
-                ActorNumber.Create(wholesaleSeries.ChargeOwnerId),
+                wholesaleSeries.ChargeOwnerId == string.Empty ? null : ActorNumber.Create(wholesaleSeries.ChargeOwnerId),
                 ActorNumber.Create(wholesaleSeries.EnergySupplierId),
                 wholesaleSeries.GridArea,
-                wholesaleSeries.ChargeCode,
+                // Protocol buffers has no concept of null, so we need to check if the string is empty
+                wholesaleSeries.ChargeCode == string.Empty ? null : wholesaleSeries.ChargeCode,
```
MWO1024, line 62: I think we should use the "has value" method instead (or whatever the name is. The one you get from proto on optional fields). It will support the case where the id or code is in fact empty string, and it is easier to read.

MWO1024 (approved):

source/Process.Application/Transactions/WholesaleServices/WholesaleServicesRequestAcceptedMapper.cs
MadsDue: Good catch :)

source/IntegrationTests/EventBuilders/WholesaleServicesResponseEventBuilder.cs
```diff
@@ -135,6 +175,10 @@ public static WholesaleServicesRequestRejected GenerateRejectedFrom(WholesaleSer
 
     private static List<WholesaleServicesRequestSeries.Types.Point> CreatePoints(WholesaleServicesRequestSeries.Types.Resolution resolution, Instant periodStart, Instant periodEnd)
     {
+        // TODO: this look wrong,
+        // if resolution is monthly, we should only have one point = a mount amount
+        // if resolution is day should be one point per day in the calculated month.
+        // Will look at this in a follow-up PR
```
ebbeknudsen, line 181: Isn't this what the following code does, or am i misunderstanding something?

source/Process.Application/Transactions/WholesaleServices/WholesaleServicesRequestAcceptedMapper.cs
```diff
@@ -47,18 +47,19 @@ public Task<INotification> MapFromAsync(byte[] payload, EventId eventId, Guid re
...
+                MapPoints(wholesaleSeries.TimeSeriesPoints, wholesaleSeries.Resolution),
                 MapMeteringPointType(wholesaleSeries),
                 MapResolution(wholesaleSeries.Resolution),
                 MapChargeType(wholesaleSeries.ChargeType),
                 MapMeasurementUnit(wholesaleSeries.QuantityUnit),
                 SettlementVersion: MapSettlementVersion(wholesaleSeries.CalculationType),
                 MapSettlementMethod(wholesaleSeries.SettlementMethod),
                 MapCurrency(wholesaleSeries.Currency),
-                ActorNumber.Create(wholesaleSeries.ChargeOwnerId),
+                wholesaleSeries.ChargeOwnerId == string.Empty ? null : ActorNumber.Create(wholesaleSeries.ChargeOwnerId),
```
ebbeknudsen, line 58: Use `HasValue` from the protobuf framework instead

source/Process.Application/Transactions/WholesaleServices/WholesaleServicesRequestAcceptedMapper.cs
ebbeknudsen, line 62: Use `HasValue` from the protobuf framework instead

source/Process.Application/Transactions/WholesaleServices/WholesaleServicesRequestAcceptedMapper.cs
```diff
@@ -123,19 +124,19 @@ private static MeasurementUnit MapMeasurementUnit(WholesaleServicesRequestSeries
         {
             WholesaleServicesRequestSeries.Types.QuantityUnit.Kwh => MeasurementUnit.Kwh,
             WholesaleServicesRequestSeries.Types.QuantityUnit.Pieces => MeasurementUnit.Pieces,
-            WholesaleServicesRequestSeries.Types.QuantityUnit.Unspecified => throw new InvalidOperationException("Could not map quantity unit"),
+            WholesaleServicesRequestSeries.Types.QuantityUnit.Unspecified => MeasurementUnit.Kwh,
```
ebbeknudsen, line 127: Is this correct? AFAIK the `Unspecified` enum value in protobuf should never be used, thus we should throw an exception if we receive one. It is not intended for a default value or something like that (https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum)

source/Process.Application/Transactions/WholesaleServices/WholesaleServicesRequestAcceptedMapper.cs
```diff
@@ -123,19 +124,19 @@ private static MeasurementUnit MapMeasurementUnit(WholesaleServicesRequestSeries
...
-    private static ChargeType MapChargeType(WholesaleServicesRequestSeries.Types.ChargeType chargeType)
+    private static ChargeType? MapChargeType(WholesaleServicesRequestSeries.Types.ChargeType chargeType)
     {
         return chargeType switch
         {
             WholesaleServicesRequestSeries.Types.ChargeType.Fee => ChargeType.Fee,
             WholesaleServicesRequestSeries.Types.ChargeType.Tariff => ChargeType.Tariff,
             WholesaleServicesRequestSeries.Types.ChargeType.Subscription => ChargeType.Subscription,
-            WholesaleServicesRequestSeries.Types.ChargeType.Unspecified => throw new InvalidOperationException("Could not map charge type"),
+            WholesaleServicesRequestSeries.Types.ChargeType.Unspecified => null,
```
ebbeknudsen, line 139: See previous comment about `Unspecified` in enums

source/IntegrationTests/EventBuilders/WholesaleServicesResponseEventBuilder.cs
```diff
@@ -48,52 +48,92 @@ public static WholesaleServicesRequestAccepted GenerateAcceptedFrom(WholesaleSer
...
-
-                var points = CreatePoints(resolution, periodStart, periodEnd);
-
-                var series = new WholesaleServicesRequestSeries()
+                var series = chargeTypes.Select(
+                    ct =>
+                    {
+                        var resolution = request.Resolution == DataHubNames.Resolution.Monthly
+                            ? WholesaleServicesRequestSeries.Types.Resolution.Monthly
+                            : WholesaleServicesRequestSeries.Types.Resolution.Hour;
```
ebbeknudsen, line 67: Can't this be `Daily` as well?

PR #1150, opened 2024-08-08
fix: quantity quality should not be part of any montly amounts

MadsDue:
## Description
The business requirements for writing the QuantityQuality have been align/updated. For any monthly amounts (including total amount) the field should be removed and not set as calculated anymore. 

https://energinet.atlassian.net/wiki/spaces/DHDOCS/pages/939950357/RSM-019+Fremsendelse+af+beregnede+engrosydelser


source/IntegrationTests/DocumentAsserters/NotifyWholesaleServicesDocumentAsserter.cs
```diff
@@ -90,14 +90,12 @@ public static async Task AssertCorrectDocumentAsync(DocumentFormat documentForma
         var isTotalMonthlyAmount = assertionInput.Points.First().Price is null;
         if (!isTotalMonthlyAmount)
```
lasrinnil, line 91:
```suggestion
        if (!isTotalOrMonthlyAmount)
```
Isn't this correct? It should not be present with either total or montly sums

lasrinnil: Maybe this change is incorrect. But it is hard to determine if this change is only for the Monthly amount messages or for both Total and Monthly amount messages

lasrinnil (approved):

source/IntegrationTests/DocumentAsserters/NotifyWholesaleServicesDocumentAsserter.cs
MadsDue: this should only happen for TotalMonthlyAmount and not MonthlyAmount because MonthlyAmount has more fields on the point to validate

PR #1151, opened 2024-08-09
fix: using unspecified goes against protobuf recommendations

MadsDue:


lasrinnil (approved):
Maybe we should have a unit test to verify behavior is as expected.

PR #1157, opened 2024-08-13
fix: ensure price measure unit is align with business requirements

MadsDue:
## Description

Price measure unit for Subscription and fee should be pieces  and the field should not be present on total monthly sum (both request and calculation results)

## References
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/276


source/BuildingBlocks.Domain/Models/MeasurementUnit.cs
```diff
@@ -32,4 +31,23 @@ private MeasurementUnit(string name, string code)
...
+    public static MeasurementUnit? TryFromChargeType(ChargeType? chargeType)
+    {
+        if (chargeType is null) return null;
+
+        var chargeTypeToMeasurementUnitMap = new Dictionary<ChargeType, MeasurementUnit>
+        {
+            { ChargeType.Tariff, Kwh },
+            { ChargeType.Subscription, Pieces },
+            { ChargeType.Fee, Pieces },
+        };
```
MWO1024, line 44: It would be cool if there was some place nice to place this. I have no idea where I would put it though, so I'm ok with leaving it as it is. It just seems like an important piece of static business logic.

MWO1024 (approved):

source/BuildingBlocks.Domain/Models/MeasurementUnit.cs
MadsDue: I saw this as some domain logic for the measure unit just like we do in TransactionId :)

MWO1024: Let keep it as it is.

PR #1162, opened 2024-08-15
feat: added new idempotent id field to outgoing message

MadsDue:
## Description
Since a calculation result no longer corresponds to one outgoing message we can't use the external id to check for idempotency. Instead, an OutgoingMessageIdempotentId has been introduced in this PR and OutgoingMessage now requires an idempotent field when it's being initialized.


source/Process.Application/Transactions/WholesaleServices/Commands/Handlers/RejectProcessWhenRejectedWholesaleServicesIsAvailable.cs
```diff
@@ -23,18 +23,23 @@
...
+    private readonly IClock _clock;
 
     public RejectProcessWhenRejectedWholesaleServicesIsAvailable(
-        IWholesaleServicesProcessRepository wholesaleServicesProcessRepository)
+        IWholesaleServicesProcessRepository wholesaleServicesProcessRepository,
+        IClock clock)
     {
         _wholesaleServicesProcessRepository = wholesaleServicesProcessRepository;
+        _clock = clock;
     }
```
MWO1024, line 43: How is the clock used?

MadsDue: It is not :)

MWO1024 (dismissed):

MWO1024 (approved):

source/ApplyDBMigrationsApp/Scripts/Model/202408141237 Add IdempotentId to OutgoingMessages.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE [dbo].[OutgoingMessages]
+    ADD [IdempotentId] INT NOT NULL DEFAULT(0);
+GO
+
+-- recreate the index with the new IdempotentId column
+DROP INDEX [dbo].[OutgoingMessages].IDX_OutgoingMessageIdempotency;
+CREATE UNIQUE INDEX IDX_OutgoingMessageIdempotency ON [dbo].[OutgoingMessages] (IdempotentId);
```
lasrinnil, line 7: Do we still need an index on ExternalId if this is used to check if a message have been created?

source/OutgoingMessages.Application/OutgoingMessageFactory.cs
```diff
@@ -51,7 +52,10 @@ public static OutgoingMessage CreateMessage(
             gridAreaCode: acceptedMessage.Series.GridAreaCode,
             externalId: acceptedMessage.ExternalId,
             calculationId: null,
-            documentReceiver: Receiver.Create(acceptedMessage.DocumentReceiverNumber, acceptedMessage.DocumentReceiverRole));
+            OutgoingMessageIdempotentId.New(
```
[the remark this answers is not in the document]
MadsDue: Since the ExternalId is created per requests, it should be fine :)

source/ApplyDBMigrationsApp/Scripts/Model/202408141237 Add IdempotentId to OutgoingMessages.sql
MadsDue: It has been refactored in this PR to look at the IdempotentId field to determine if a message has already been created.

lasrinnil (dismissed):

lasrinnil (approved):

PR #1164, opened 2024-08-16
feat: split wholesale result set if a expected point is missing

MadsDue:


source/OutgoingMessages.Infrastructure/Databricks/Factories/PeriodFactory.cs
```diff
@@ -29,19 +29,23 @@ public static class PeriodFactory
     /// The oldest point is the start of the calculation period.
     /// The newest point plus the resolution is the end of the calculation period.
     /// </summary>
-    /// <param name="timeSeriesPoints"></param>
-    /// <param name="resolution"></param>
-    public static Period GetPeriod(IReadOnlyCollection<WholesaleTimeSeriesPoint> timeSeriesPoints, Resolution resolution)
+    public static Period GetPeriod(
+        IReadOnlyCollection<WholesaleTimeSeriesPoint> timeSeriesPoints,
+        Resolution resolution,
+        DateTimeZone dateTimeZone)
```
lasrinnil, line 35: I would suggest that we refactor `PeriodFactory` to no longer be `static` this way it could have dependency to the `DateTimeZone` without having to have every activity have the dependency for the sake of forwarding it to the `PeriodFactory`.

source/OutgoingMessages.Infrastructure/Databricks/WholesaleResults/Queries/WholesaleResultQueryBase.cs
```diff
@@ -129,14 +127,30 @@ private static WholesaleTimeSeriesPoint CreateTimeSeriesPoint(DatabricksSqlRow d
...
+    /// <summary>
+    /// Checks if the current result follows the previous result based on time and resolution.
+    /// </summary>
+    private bool IsNextInResultSequence(DatabricksSqlRow currentResult, DatabricksSqlRow previousResult)
     {
-        return currentResultId == rowResultId;
+        var endTimeOfPreviousResult = GetEndTimeOfPreviousResult(previousResult);
+
+        return endTimeOfPreviousResult == currentResult.ToInstant(WholesaleResultColumnNames.Time);
+    }
```
lasrinnil, line 145:
```
    /// <summary>
    /// Checks if the current result follows the previous result based on time and resolution.
    /// </summary>
    private bool ResultStartEqualsPreviousResultEnd(DatabricksSqlRow result, DatabricksSqlRow previousResult)
    {
        var endTimeFromPreviousResult = GetEndTimeOfPreviousResult(previousResult);
        var startTimeFromResult = result.ToInstant(WholesaleResultColumnNames.Time);

        // The start time of the current result should be the same as the end time of the previous result if the result is in sequence with the previous result.
        return endTimeFromPreviousResult == startTimeFromResult;
    }
```

lasrinnil: I might just be me, but I had a hard time grasping what `IsNextInResultSequence` did initially. Im going for a more explicit approach, I just might be overdoing it.

lasrinnil (approved):
Good stuff man!

lasrinnil (approved):

PR #1165, opened 2024-08-16
feat: split energy result set if a gap in point is detected

MadsDue:
## Description
Messages are only allowed to contain points in a time-based sequence based on their resolution. If we detect a missing point in a result set, we now split the message into multiple messages without missing points.


lasrinnil (approved):
It would be great if we could creating a base for both Energy and Wholesale results from which they could call: `  internal async IAsyncEnumerable<QueryResult<TResult>> GetAsync(DatabricksSqlWarehouseQueryExecutor databricksSqlWarehouseQueryExecutor)` in another pr 👍

lasrinnil (approved):

lasrinnil (approved):

PR #1166, opened 2024-08-16
docs: Reduce dependencies in C4

dstenroejl:
## Description

Wholesale doesn't link to to Market Participant and Revision Log models anymore, which means we can also remove them from EDI and make C4 simpler.

## References

https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/217


ebbeknudsen (approved):
LGTM

PR #1184, opened 2024-08-27
chore: Bump NuGet packages

dstenroejl:
## Description

- [x] Bump NuGet packages
- [x] Register "status" endpoint for heath checks

## Checklist
- [ ] Reference to the task

### DIFF
--- a/source/B2CWebApi/Program.cs
+++ b/source/B2CWebApi/Program.cs
@@ -96,8 +96,8 @@ app
 app.MapControllers().RequireAuthorization();
 
 app.MapLiveHealthChecks();
-
 app.MapReadyHealthChecks();
+app.MapStatusHealthChecks();
 
 app.Run();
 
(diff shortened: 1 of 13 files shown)


source/B2CWebApi/Program.cs
```diff
@@ -96,8 +96,8 @@
 app.MapControllers().RequireAuthorization();
 
 app.MapLiveHealthChecks();
-
 app.MapReadyHealthChecks();
+app.MapStatusHealthChecks();
```
ebbeknudsen, line 100: Does this need to be done in the B2B function app as well?

ebbeknudsen (approved):
LGTM

source/B2CWebApi/Program.cs
dstenroejl: No.

PR #1185, opened 2024-08-29
fix: ensure grid operators can request energy results from b2c

MadsDue:
## Description
Grid operators should be able to request energy result from b2c channel because of the MDR/DDM hack. 

## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/10627883955
- [ ] Reference to the task


MWO1024 (approved):

source/IntegrationTests/Behaviours/IncomingRequests/GivenB2CAggregatedMeasureDataRequestTests.cs
```diff
@@ -51,6 +51,8 @@ public static object[][] DocumentFormatsWithMarketRoleCombinations()
             MarketRole.EnergySupplier,
             MarketRole.BalanceResponsibleParty,
             MarketRole.MeteredDataResponsible,
```
ebbeknudsen, line 53: Use `WorkaroundFlags.MeteredDataResponsibleToGridOperatorHack` so we easily can find the places the "hack" is in place, and easily remove the code when we no longer want it :)

PR #1202, opened 2024-09-04
refac: Use ServiceBus IAM

dstenroejl:
## Description

Refactor any use of ServiceBus to identity access management (IAM) instead of shared access policies (connection strings).

**Important changes:**
- When using `ServiceBusTrigger` we must configure a setting with the name `<SectionName>_FullyQualifiedNamespace` and configure the trigger parameter `Connection` with the `<SectionName>` value.
- When configuring `AddAzureClients()` we must use the `UseCredential()` and `AddServiceBusClientWithNamespace()` to configure the top level `ServiceBusClient`. Any place where we need a service bus sender/receiver/processor we must inject `ServiceBusClient` and use it to create the lower level clients.
- When configuring health checks we must use the overloads where we specify the parameters `fullyQualifiedNamespaceFactory` and `tokenCredentialFactory`. Set `tokenCredentialFactory` to `new DefaultAzureCredential()` - which isn't costly as under the hood the health checks will cache the `ServiceBusClient` using the `FullyQualifiedNamespace` as key.

Deployed to `dev_002`: https://github.com/Energinet-DataHub/dh3-environments/actions/runs/10703545430

## References

https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/293

## Checklist
- [x] Reference to the task

### DIFF
--- a/source/ArchitectureTests/RegistrationTests.cs
+++ b/source/ArchitectureTests/RegistrationTests.cs
@@ -57,9 +57,7 @@ public class RegistrationTests
         // The following declaration slows down the test execution, since creating a new Uri is a heavy operation
         Environment.SetEnvironmentVariable("AZURE_STORAGE_ACCOUNT_URL", TestEnvironment.CreateFakeStorageUrl());
 
-        Environment.SetEnvironmentVariable($"{ServiceBusOptions.SectionName}__{nameof(ServiceBusOptions.ManageConnectionString)}", TestEnvironment.CreateFakeServiceBusConnectionString());
-        Environment.SetEnvironmentVariable($"{ServiceBusOptions.SectionName}__{nameof(ServiceBusOptions.ListenConnectionString)}", TestEnvironment.CreateFakeServiceBusConnectionString());
-        Environment.SetEnvironmentVariable($"{ServiceBusOptions.SectionName}__{nameof(ServiceBusOptions.SendConnectionString)}", TestEnvironment.CreateFakeServiceBusConnectionString());
+        Environment.SetEnvironmentVariable($"{ServiceBusOptions.SectionName}__{nameof(ServiceBusOptions.FullyQualifiedNamespace)}", TestEnvironment.CreateFakeServiceBusFullyQualifiedNamespace());
 
         Environment.SetEnvironmentVariable(nameof(DatabricksSqlStatementOptions.WorkspaceUrl), "https://adb-1000.azuredatabricks.net/");
         Environment.SetEnvironmentVariable(nameof(DatabricksSqlStatementOptions.WorkspaceToken), "FakeToken");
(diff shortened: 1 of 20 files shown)


source/ArchitectureTests/RegistrationTests.cs
```diff
@@ -279,12 +275,10 @@ private sealed class TestEnvironment : RuntimeEnvironment
...
         public override string AZURE_FUNCTIONS_ENVIRONMENT => "Development";
 
-        public static string CreateFakeServiceBusConnectionString()
+        public static string CreateFakeServiceBusFullyQualifiedNamespace()
         {
             return new StringBuilder()
-                .Append(CultureInfo.InvariantCulture, $"Endpoint=sb://sb-{Guid.NewGuid():N}.servicebus.windows.net/;")
-                .Append("SharedAccessKeyName=send;")
-                .Append(CultureInfo.InvariantCulture, $"SharedAccessKey={Guid.NewGuid():N}")
+                .Append(CultureInfo.InvariantCulture, $"sb://sb-{Guid.NewGuid():N}.servicebus.windows.net/")
```
ebbeknudsen, line 281: I don't think this makes sense as a `StringBuilder`, but you don't have to change it, since it works perfectly fine like this

source/BuildingBlocks.Application/Extensions/Options/ServiceBusOptions.cs
```diff
@@ -16,16 +16,13 @@
...
 
     [Required]
-    public string ManageConnectionString { get; set; } = string.Empty;
-
-    [Required]
-    public string ListenConnectionString { get; set; } = string.Empty;
-
-    [Required]
-    public string SendConnectionString { get; set; } = string.Empty;
+    public string FullyQualifiedNamespace { get; set; } = string.Empty;
```
ebbeknudsen, line 27: I think there might be sample config files that needs to be updated as well (and the dev-secrets repository)

source/ArchitectureTests/RegistrationTests.cs
dstenroejl: I don't want to make additional changes, I didn't write that code.

source/B2BApi/EventListeners/InboxEventListener.cs
```diff
@@ -50,7 +50,7 @@ public InboxEventListener(
     public async Task RunAsync(
         [ServiceBusTrigger(
             $"%{EdiInboxOptions.SectionName}:{nameof(EdiInboxOptions.QueueName)}%",
-            Connection = $"{ServiceBusOptions.SectionName}:{nameof(ServiceBusOptions.ManageConnectionString)}")]
+            Connection = ServiceBusOptions.SectionName)]
```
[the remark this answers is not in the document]
dstenroejl: Its Microsoft, not sure what to say. ;)
But you can read about it in their documentation, under ServiceBusTrigger.

source/BuildingBlocks.Application/Extensions/Options/ServiceBusOptions.cs
dstenroejl: Yes, I have a task for this in the issue that I reference in the PR.

PR #1203, opened 2024-09-05
feat: Impl duration metric tracking of message enqueueing

MadsDue:
## Description
To monitor the execution of enqueue messages, we now track the duration of the enqueue messages orchestrator through Application insight metrics.

E.g.
image

## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/workflows/edi-cd.yml
- [x] Reference to the task => https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/245


source/B2BApi/B2BApi.csproj
```diff
@@ -24,6 +24,7 @@ limitations under the License.
     </PropertyGroup>
     <ItemGroup>
         <PackageReference Include="Energinet.DataHub.Core.Databricks.SqlStatementExecution" Version="11.1.0" />
+        <PackageReference Include="Microsoft.ApplicationInsights" Version="2.22.0" />
```
lasrinnil, line 27: I would actually expected we would have dependency to this package from the "App" package. But lets got with it for now (y)

lasrinnil (approved):

PR #1205, opened 2024-09-06
refac: DDM request rsm16 as MDR hack

lasrinnil:
## Checklist
- [ ] Reference to the task

### DIFF
--- a/source/B2CWebApi/Factories/RequestAggregatedMeasureDataDtoFactory.cs
+++ b/source/B2CWebApi/Factories/RequestAggregatedMeasureDataDtoFactory.cs
@@ -138,11 +138,10 @@ public static class RequestAggregatedMeasureDataDtoFactory
             return marketRole.Code;
         }
 
-        if (WorkaroundFlags.MeteredDataResponsibleToGridOperatorHack
-            && marketRole == MarketRole.GridAccessProvider
-            && marketRole.Code != null)
+        if (WorkaroundFlags.GridOperatorToMeteredDataResponsibleHack // TODO: LRN Where do we test this?
+            && marketRole == MarketRole.GridAccessProvider)
         {
-            return marketRole.Code;
+            return MarketRole.MeteredDataResponsible.Code!;
         }
 
         throw new ArgumentException($"Market Role: {marketRole}, is not allowed to request aggregated measure data.");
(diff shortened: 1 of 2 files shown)


source/Tests/B2CWebApi/RequestAggregatedMeasureData/RequestAggregatedMeasureDataFactoryTests.cs
```diff
@@ -89,6 +76,34 @@ public void Can_create_RequestAggregatedMeasureData_wrong_endData_input()
...
 
         var endDate = InstantPattern.General.Parse(result.Serie.First().EndDateAndOrTimeDateTime!).GetValueOrThrow();
-        Assert.Equal(endDay - 1, endDate.Day());
+        Assert.Equal(EndDay - 1, endDate.Day());
+    }
+
+    [Fact]
+    public void RequestingRequestAggregatedMeasureData_AsGridOperator_RequestingRoleIsChangedToMeteredDataResponsible()
+    {
+        var request = RequestAggregatedMeasureDataMarketRequest();
```
ebbeknudsen, line 85: Consider a summary describing that the test verifies the `WorkaroundFlags.GridOperatorToMeteredDataResponsibleHack`.

We could maybe even start the test with something like:
```c#
if (!WorkaroundFlags.GridOperatorToMeteredDataResponsibleHack)
    throw new Exception("Remove this test when GridOperatorToMeteredDataResponsibleHack is no longer in use");
```

Then the `WorkaroundFlags.GridOperatorToMeteredDataResponsibleHack` can be used to find tests that should be removed when the hack is removed as well.

source/IntegrationTests/Behaviours/IncomingRequests/GivenB2CAggregatedMeasureDataRequestTests.cs
```diff
@@ -49,8 +49,6 @@ public static object[][] DocumentFormatsWithMarketRoleCombinations()
             MarketRole.EnergySupplier,
             MarketRole.BalanceResponsibleParty,
             MarketRole.MeteredDataResponsible,
-            // GridAccessProvider is allowed to act as a MeteredDataResponsible (MeteredDataResponsibleToGridOperatorHack)
-            MarketRole.GridAccessProvider,
```
ebbeknudsen, line 53: Isn't a `GridAccessProvider` still allowed to make B2C requests? The `RequestedForActorRole` in the tests is then expected to be `MDR` instead of `DDM`, since we change it in the factory, or have i misunderstood?

lasrinnil: Updated the test to reflect your feedback 👍

PR #1208, opened 2024-09-09
refac: Remove DbContext execution strategy

ebbeknudsen:
## Description

Remove DbContext execeution strategy, since it requires transactions to set an execution strategy manually, which introduces bugs when having transactions across multiple database contexts.

- Remove `EnableRetryOnFailure` from db contexts
- Remove manually setting execution strategy from `ResillientTransaction`
- Update audit log tests to check for request success

## References

- [x] d002 run: https://github.com/Energinet-DataHub/dh3-environments/actions/runs/10770640439/job/29864433631

## Checklist
- [ ] Reference to the task


source/B2CWebApi.AppTests/Tests/B2CWebApiAuditLogTests.cs
```diff
@@ -31,6 +31,10 @@ namespace Energinet.DataHub.EDI.B2CWebApi.AppTests.Tests;
 [Collection(nameof(B2CWebApiCollectionFixture))]
 public class B2CWebApiAuditLogTests : IAsyncLifetime
 {
+    private static readonly string _datahubAdministratorRole = "DataHubAdministrator";
```
lasrinnil, line 34: Why is this test included in this PR? Is it relevant?

lasrinnil (approved):

PR #1211, opened 2024-09-09
refac: Use Azure SDK extensions for ServiceBus configuration

dstenroejl:
### Registrations

Refactored the creation of ServiceBus clients to follow best practices mentioned here:
* https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-performance-improvements?tabs=net-standard-sdk-2#reusing-factories-and-clients
* https://learn.microsoft.com/en-us/dotnet/azure/sdk/dependency-injection?tabs=host-builder#register-clients-and-subclients

By using the Azure SDK extensions for dependency injection we get the following benefits:
- Singleton instances
  - Lazy creation of instances
  - Cached instances
- Since instances are registered to the DI container we don't have to manually dispose them

### ServiceBus adaption and tests

Deleted custom factory `ServiceBusSenderFactory` and sender adapter `ServiceBusSenderAdapter`. This design was implemented to be able to test communication without a servicebus, but that is still possible as can be seen by looking at the refactored `ServiceBusSenderFactoryStub` and `ServiceBusSenderSpy`.

## References

https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/293

## Checklist
  - Deployed to `dev_002`: https://github.com/Energinet-DataHub/dh3-environments/actions/runs/10775630118
- [x] Reference to the task


source/IncomingMessages.Application/Extensions/DependencyInjection/IncomingMessagesExtensions.cs
```diff
@@ -78,18 +72,50 @@ public static IServiceCollection AddIncomingMessagesModule(this IServiceCollecti
...
+                .WithName(incommingMessagesQueueOptions.QueueName);
+        });
+
+        // => Health checks
+        services.AddHealthChecks()
+            .AddAzureServiceBusQueue(
+                sp => sp.GetRequiredService<IOptions<ServiceBusNamespaceOptions>>().Value.FullyQualifiedNamespace,
+                sp => sp.GetRequiredService<IOptions<IncomingMessagesQueueOptions>>().Value.QueueName,
+                _ => new DefaultAzureCredential(),
+                name: incommingMessagesQueueOptions.QueueName);
```
MWO1024, line 109: The code was changes from a 'try add' to a plain add. Are we sure this won't give any issues?

source/IntegrationTests/Application/IncomingMessages/RequestAggregatedMeasureData/InitializeAggregatedMeasureDataProcessesCommandTests.cs
```diff
@@ -35,14 +36,16 @@ namespace Energinet.DataHub.EDI.IntegrationTests.Application.IncomingMessages.Re
 public class InitializeAggregatedMeasureDataProcessesCommandTests : TestBase
 {
     private readonly ProcessContext _processContext;
+#pragma warning disable CA2213 // Disposable fields should be disposed
     private readonly ServiceBusSenderSpy _senderSpy;
+#pragma warning restore CA2213 // Disposable fields should be disposed
     private readonly ServiceBusSenderFactoryStub _serviceBusClientSenderFactory;
```
MWO1024, line 42: Is this to let the framework handle the dispose?

source/Process.Application/Extensions/DependencyInjection/ProcessExtensions.cs
```diff
@@ -50,34 +50,29 @@ public static IServiceCollection AddProcessModule(this IServiceCollection servic
...
-            .AddOptionsWithValidateOnStart<EdiInboxOptions>()
-            .BindConfiguration(EdiInboxOptions.SectionName)
-            .ValidateDataAnnotations();
-
-        services
-            .AddOptions<WholesaleInboxOptions>()
-            .BindConfiguration(WholesaleInboxOptions.SectionName)
+            .AddOptionsWithValidateOnStart<EdiInboxQueueOptions>()
+            .BindConfiguration(EdiInboxQueueOptions.SectionName)
             .ValidateDataAnnotations();
```
MWO1024, line 60: I believe the process modules depends on the WholesaleInbox, so it feel weird to remove it from here. But the dependency might have changed with the EMO introduction and such, so I am no longer entirely sure.

source/Process.Application/Transactions/AggregatedMeasureData/Notifications/Handlers/NotifyWholesaleThatAggregatedMeasureDataIsRequestedHandler.cs
```diff
@@ -22,10 +22,10 @@ public sealed class
     NotifyWholesaleThatAggregatedMeasureDataIsRequestedHandler : INotificationHandler<
     NotifyWholesaleThatAggregatedMeasureDataIsRequested>
 {
-    private readonly WholesaleInbox _wholesaleInbox;
+    private readonly WholesaleInboxClient _wholesaleInbox;
```
MWO1024, line 25: `_inboxClient` or `_wholesaleInboxClient`?

source/Process.Application/Transactions/WholesaleServices/Notifications/Handlers/NotifyWholesaleThatWholesaleServicesIsRequestedHandler.cs
```diff
@@ -23,10 +23,10 @@ namespace Energinet.DataHub.EDI.Process.Application.Transactions.WholesaleServic
 
 public class NotifyWholesaleThatWholesaleServicesIsRequestedHandler : INotificationHandler<NotifyWholesaleThatWholesaleServicesIsRequested>
 {
-    private readonly WholesaleInbox _wholesaleInbox;
+    private readonly WholesaleInboxClient _wholesaleInbox;
```
MWO1024, line 26: `_inboxClient` or `_wholesaleInboxClient`?

MWO1024 (approved):

source/IncomingMessages.Application/Extensions/DependencyInjection/IncomingMessagesExtensions.cs
dstenroejl: `TryAdd` is only necessary when we have registrations that is performed from multiple places. E.g. two modules both uses the same ServiceBus queue and each module performs the health check registration.

The "IncommingMessagesQueue" health check is only registered in the method `AddIncomingMessagesModule`, and `AddIncomingMessagesModule` is only called once in each application, so we don't need `TryAdd`.

source/IntegrationTests/Application/IncomingMessages/RequestAggregatedMeasureData/InitializeAggregatedMeasureDataProcessesCommandTests.cs
dstenroejl: No. Actually there is nothing to dispose, the "Spy" is not a full implementation of a "ServiceBusSender". If we did call Dispose, we would get a null exception.

source/Process.Application/Transactions/AggregatedMeasureData/Notifications/Handlers/NotifyWholesaleThatAggregatedMeasureDataIsRequestedHandler.cs
dstenroejl: I prefer addding `Client`

MWO1024: `Client` works too.

source/Process.Application/Transactions/WholesaleServices/Notifications/Handlers/NotifyWholesaleThatWholesaleServicesIsRequestedHandler.cs
dstenroejl: I prefer addding `Client`

PR #1223, opened 2024-09-12
feat: log when delegation is performed or Charge was a tax

lasrinnil:
## Description
This PR will add additional logning when
- Message is delegated to a Actor
- Message is sent to a grid area owner.

## References
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/229

## Checklist
- [x] Reference to the task


source/MasterData.Interfaces/Models/GridAreaOwnerDto.cs
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Energinet.DataHub.EDI.BuildingBlocks.Domain.Models;
+using NodaTime;
+
+namespace Energinet.DataHub.EDI.MasterData.Interfaces.Models;
+
+public record GridAreaOwnerDto(string GridAreaCode, Instant ValidFrom, ActorNumber ActorNumber, int SequenceNumber);
```
MadsDue, line 20: Is this a GridAreaOwner? 

It should a bit more like a GridAreaOwnership

MadsDue: Instant ValidFrom
int SequenceNumber

Not sure why we want to expose those two fields. Aren't they just a concern for the domain model?

source/OutgoingMessages.Application/UseCases/DelegateMessage.cs
```diff
@@ -83,6 +83,16 @@ public async Task<OutgoingMessage> DelegateAsync(
...
+                "Message was delegated from {FromActorNumber} to {ToActorNumber}, with role {ActorRole} based on delegation with sequence number {SequenceNumber}",
+                delegatedByActorNumber.Value,
+                messageDelegation.DelegatedTo.ActorNumber.Value,
+                delegatedByActorRole.Code,
+                messageDelegation.SequenceNumber);
+        }
+
         return messageDelegation is not null
             ? Receiver.Create(messageDelegation.DelegatedTo.ActorNumber, messageDelegation.DelegatedTo.ActorRole)
             : null;
```
MadsDue, line 98: Could we log the receiver instead? 
- Receiver.Create() could potentiel do should else.

```suggestion
var receiver = messageDelegation is not null
            ? Receiver.Create(messageDelegation.DelegatedTo.ActorNumber, messageDelegation.DelegatedTo.ActorRole)
            : null
        if (receiver != null)
        {
            _logger.LogInformation(
                "Message was delegated from {FromActorNumber} to {ToActorNumber}, with role {ActorRole} based on delegation with sequence number {SequenceNumber}",
                delegatedByActorNumber.Value,
                receiver.ActorNumber.Value,
                 receiver.ActorRode.Name,
                messageDelegation.SequenceNumber);
        }

        return receiver;
```

source/MasterData.Interfaces/Models/GridAreaOwnerDto.cs
lasrinnil: `ValidFrom` is just for now a concern for the domain model. But the `SequenceNumber` is used when logging that the GridAreaOwner ActorNumber is being used. If we had a EventId we could have perhaps chosen this instead.  But this way we log with an identifier which could be used to track the information back to `MarketParticipant` subsystem.

lasrinnil: > Is this a GridAreaOwner?
> 
> It should a bit more like a GridAreaOwnership

Whats the difference in this case? In daily talk we use `GridAreaOwner` imo.

source/OutgoingMessages.Application/UseCases/DelegateMessage.cs
lasrinnil: please see my last commit 👍

MadsDue (approved):

PR #1238, opened 2024-09-17
fix: auth actor should be sender on incoming archived messages

MadsDue:
## Description
To ensure that an aktør can be associated with its archived messages. We should use the sender number from the authenticated actor instead of the message content. 

Following PRs:
- Persists SenderRole and ReceiverRole on an archive message
- Add authenticated actor role filter if the actor has 'own data' restriction

## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/10898575813
- [x] Reference to the task => https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/146


source/IncomingMessages.Application/UseCases/ReceiveIncomingMarketMessage.cs
```diff
@@ -131,11 +135,13 @@ private async Task ArchiveIncomingMessageAsync(
         IncomingDocumentType incomingDocumentType,
         CancellationToken cancellationToken)
     {
+        Thread.Sleep(5000);
+        var authenticatedActor = _actorAuthenticator.CurrentActorIdentity;
         await _archivedMessagesClient.CreateAsync(
```
MWO1024, line 140: Why the sleep?

MWO1024 (dismissed):

source/IncomingMessages.Application/UseCases/ReceiveIncomingMarketMessage.cs
MadsDue: haha, good catch. it was for testing

MWO1024 (approved):

PR #1242, opened 2024-09-17
feat: Use latin supplement and extended-A in json encoding

MWO1024:
## Description

Update the JSON encoding to support both latin supllement and lation extended-A unicode ranges. These should gives us enough special characters for our need, including 'æ', 'ø', and 'å'.

## References

The Latin-1 Supplement (also called C1 Controls and Latin-1 Supplement) is the second [Unicode block](https://en.wikipedia.org/wiki/Unicode_block) in the [Unicode](https://en.wikipedia.org/wiki/Unicode) standard. It encodes the upper range of [ISO 8859-1](https://en.wikipedia.org/wiki/ISO_8859-1): 80 (U+0080) - FF (U+00FF). [C1 Controls](https://en.wikipedia.org/wiki/C0_and_C1_control_codes) (0080–009F) are not graphic. This block ranges from U+0080 to U+00FF, contains 128 characters and includes the [C1 controls](https://en.wikipedia.org/wiki/C1_controls), [Latin-1](https://en.wikipedia.org/wiki/Latin-1) [punctuation](https://en.wikipedia.org/wiki/Punctuation) and [symbols](https://en.wikipedia.org/wiki/Symbols), 30 pairs of [majuscule](https://en.wikipedia.org/wiki/Majuscule) and [minuscule](https://en.wikipedia.org/wiki/Lower_case) [accented](https://en.wikipedia.org/wiki/Accented) [Latin characters](https://en.wikipedia.org/wiki/Latin_characters) and 2 mathematical operators. Major alphabets included: French, German, Icelandic, Portuguese, and Spanish.

Latin Extended-A is a [Unicode block](https://en.wikipedia.org/wiki/Unicode_block) and is the third block of the Unicode standard. It encodes Latin letters from the [Latin ISO character sets](https://en.wikipedia.org/wiki/ISO/IEC_8859) other than Latin-1 (which is already encoded in the [Latin-1 Supplement](https://en.wikipedia.org/wiki/Latin-1_Supplement_(Unicode_block)) block) and also legacy characters from the [ISO 6937](https://en.wikipedia.org/wiki/ISO_6937) standard. Major alphabets included: Afrikaans, Catalan, Croatian, Czech, Esperanto, Greenlandic, Hungarian, Kashubian, Kurdish, Latin, Latvian, Lithuanian, Maltese, Northern Sami, Polish, Romanian, Serbian, Slovak, Slovene, Sorbian, Turkish, and Welsh.

https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/internal-repo/1041

## Checklist
- [x] Reference to the task


source/OutgoingMessages.Domain/DocumentWriters/NotifyAggregatedMeasureData/NotifyAggregatedMeasureDataCimJsonDocumentWriter.cs
```diff
@@ -23,20 +24,17 @@
...
+public sealed class NotifyAggregatedMeasureDataCimJsonDocumentWriter(
+    IMessageRecordParser parser,
+    JavaScriptEncoder encoder)
+    : IDocumentWriter
 {
     private const string DocumentTypeName = "NotifyAggregatedMeasureData_MarketDocument";
     private const string TypeCode = "E31";
-    private readonly IMessageRecordParser _parser;
+    private readonly IMessageRecordParser _parser = parser;
+    private readonly JsonWriterOptions _options = new() { Indented = true, Encoder = encoder };
```
lasrinnil, line 35: Could we move `Indented = true` into the `Serializer`? Would be nice if you only could overwrite the `JsonWriterOptions` but as default they were configured on the `Serializer`

source/Tests/Infrastructure/OutgoingMessages/JsonEncoderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+using Energinet.DataHub.Edi.Responses;
+using FluentAssertions;
+using Google.Protobuf;
+using Xunit;
+
+namespace Energinet.DataHub.EDI.Tests.Infrastructure.OutgoingMessages;
+
+public sealed class JsonEncoderTests
+{
+    private const string TestString = "ÅØÆÄÖåøæäö";
```
lasrinnil, line 29: é?

MWO1024: Expanded to `ÅØÆÄÖÜåøæäöüáÁéÉíÍóÓúÚýÝþðÐ`

lasrinnil (commented):
LGTM

lasrinnil (approved):
LGTM more

PR #1251, opened 2024-09-24
feat: Fix encoding and parse error in JSON requests

MWO1024:
## References

https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/223

## Checklist
  - No
  - A `git revert`-commit should do the trick
  - No changes
- [x] Reference to the task


source/B2BApi.AppTests/IncomingMessages/IncomingMessageReceiverTests.cs
```diff
@@ -97,9 +97,59 @@ public async Task Given_PersistedActor_When_CallingIncomingMessagesWithValidDocu
...
+        await Fixture.DatabaseManager.AddActorAsync(actorNumber, externalId);
+
+        // The bearer token must contain:
+        //  * the actor role matching the document content
+        //  * the external id matching the actor in the database
+        var b2bToken = new JwtBuilder()
+            .WithRole(ClaimsMap.RoleFrom(actorRole).Value)
+            .WithClaim(ClaimsMap.ActorId, externalId)
+            .CreateToken();
+
```
RasmusGraabaek, line 131: Would it be beneficial to move some of this setup to a private method. Since it is duplicated in many of the tests?

source/B2BApi/Configuration/Middleware/UnHandledExceptionMiddleware.cs
```diff
@@ -56,11 +57,33 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+
+                var contentType = httpReqData.Headers.TryGetContentType();
+
+                if (contentType is not null && contentType.Contains("application/json"))
+                {
+                    newHttpResponse.Headers.Add("Content-Type", "application/json; charset=utf-8");
+                    // You need to explicitly pass the status code in WriteAsJsonAsync method.
+                    // https://github.com/Azure/azure-functions-dotnet-worker/issues/776
+                    await newHttpResponse.WriteAsJsonAsync(
+                        new { Message = "An unexpected error occurred! Please try later." },
```
RasmusGraabaek, line 69: Can it be a const, to prevent the same magic string occurring at line 69, 78, 85?

source/IncomingMessages.Infrastructure/MessageParsers/WholesaleSettlementMessageParsers/WholesaleSettlementJsonMessageParser.cs
```diff
@@ -74,6 +70,10 @@ public async Task<IncomingMarketMessageParserResult> ParseAsync(IIncomingMarketM
         {
             return InvalidJsonFailure(e);
         }
+        catch (KeyNotFoundException e)
+        {
+            return InvalidJsonFailure(new KeyNotFoundException("Missing root element", e));
```
RasmusGraabaek, line 75: Will this only occur if the "root" element is missing?

source/Tests/CimMessageAdapter/Messages/RequestAggregatedMeasureData/MessageParserTests.cs
```diff
@@ -42,47 +43,57 @@ public class MessageParserTests
...
-                new object[] { DocumentFormat.Xml, CreateBaseXmlMessage("BadVersionRequestAggregatedMeasureData.xml"), nameof(InvalidBusinessReasonOrVersion) },
-                new object[] { DocumentFormat.Xml, CreateBaseXmlMessage("VersionIndexOutOfRangeRequestAggregatedMeasureData.xml"), nameof(InvalidMessageStructure) },
-                new object[] { DocumentFormat.Json, CreateBaseJsonMessages("FailSchemeValidationAggregatedMeasureData.json"), nameof(InvalidMessageStructure) },
-                new object[] { DocumentFormat.Json, CreateBaseJsonMessages("InvalidJsonAggregatedMeasureData.json"), nameof(InvalidMessageStructure) },
-        };
+        return
+        [
+            [
+                DocumentFormat.Xml, CreateBaseXmlMessage("BadVersionRequestAggregatedMeasureData.xml"),
+                nameof(InvalidBusinessReasonOrVersion),
```
RasmusGraabaek, line 78: Would prefer that either everything was on the same line or that we have a newline after efter "," :)

source/Tests/CimMessageAdapter/Messages/RequestAggregatedMeasureData/MessageParserTests.cs
```diff
@@ -124,30 +136,31 @@ public async Task Successfully_parsed_b2c_messages(DocumentFormat format, Stream
...
-
-        foreach (var serie in marketMessage.Series.Cast<RequestAggregatedMeasureDataMessageSeries>())
+        marketMessage.Should().NotBeNull();
+        marketMessage.MessageId.Should().Be("123564789123564789123564789123564789");
+        marketMessage.BusinessReason.Should().Be("D05");
+        marketMessage.SenderNumber.Should().Be("5799999933318");
+        marketMessage.SenderRoleCode.Should().Be("DDK");
+        marketMessage.ReceiverNumber.Should().Be("5790001330552");
+        marketMessage.ReceiverRoleCode.Should().Be("DGL");
+        // marketMessage.CreatedAt.Should().Be("2022-12-17T09:30:47Z");
```
RasmusGraabaek, line 149: We should add the "// marketMessage.CreatedAt.Should().Be("2022-12-17T09:30:47Z") " check aswell. 
Since it is being asserted in "source/Tests/CimMessageAdapter/Messages/WholesaleSettlementMessageParsers/MessageParserTests.cs"

RasmusGraabaek: Maybe in another PR. But this is header information, which is parsed by the "baseParsers"

RasmusGraabaek (approved):

source/IncomingMessages.Infrastructure/MessageParsers/WholesaleSettlementMessageParsers/WholesaleSettlementJsonMessageParser.cs
MWO1024: At the moment... probably? There isn't any compile time guarantee, and it relies solely on the premise that the schema validation will catch all missing properties, before we try and fetch them in the block where we find the root element, and it's content.
We could also change the implementation to use a dictionary, and then we would have another potential `KeyNotFoundException` there. 

We could "fix" this, by wrapping that part of the code in (yet another) try-catch block. Then we would be fairly sure, that that particular `KNFE` would be due to a missing root element.

So, in short, the answer is "no".

RasmusGraabaek (dismissed):

source/B2BApi.AppTests/TestData/Messages/xml/RequestWholesaleSettlement.xml
```diff
@@ -0,0 +1,27 @@
...
+<!--Sample XML file generated by XMLSpy v2021 rel. 3 (x64) (http://www.altova.com)-->
+<cim:RequestWholesaleSettlement_MarketDocument xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cim="urn:ediel.org:measure:requestwholesalesettlement:0:1" xsi:schemaLocation="urn:ediel.org:measure:requestwholesalesettlement:0:1 urn-ediel-org-measure-requestwholesalesettlement-0-1.xsd">
+    <cim:mRID>{MessageId}</cim:mRID>
+    <cim:type>D21</cim:type>
+    <cim:process.processType>D05</cim:process.processType>
+    <cim:businessSector.type>23</cim:businessSector.type>
+    <cim:sender_MarketParticipant.mRID codingScheme="A10">5790000392551</cim:sender_MarketParticipant.mRID>
+    <cim:sender_MarketParticipant.marketRole.type>DDQ</cim:sender_MarketParticipant.marketRole.type>
+    <cim:receiver_MarketParticipant.mRID codingScheme="A10">5790001330552</cim:receiver_MarketParticipant.mRID>
+    <cim:receiver_MarketABCParticipant.marketRole.type>DGL</cim:receiver_MarketABCParticipant.marketRole.type>
```
MWO1024, line 11: This line is deliberately wrong. It is to ensure we get a validation error in our test.

source/B2BApi/Common/HttpRequestDataExtensions.cs
```diff
@@ -60,12 +60,13 @@ private static async Task<HttpResponseData> GetContentTypeErrorResponseAsync(
         string message,
         CancellationToken cancellationToken)
     {
-        var missingContentTypeResponse = request.CreateResponse(
-            HttpStatusCode.UnsupportedMediaType);
+        var missingContentTypeResponse = request.CreateResponse(HttpStatusCode.UnsupportedMediaType);
+        missingContentTypeResponse.Headers.Add("Content-Type", "text/plain; charset=utf-8");
```
MWO1024, line 64: We don't really have a proper content type here (as that is the reason we fail). We thus default to `text/plain` and provide a simple error message. Maybe there is a better way to do it?

source/BuildingBlocks.Domain/Models/DocumentFormat.cs
```diff
@@ -30,4 +30,24 @@ public static DocumentFormat FromName(string name)
...
+
+        if (Name.Equals(Json.Name, StringComparison.OrdinalIgnoreCase))
+        {
+            return "application/json";
+        }
+
+        if (Name.Equals(Ebix.Name, StringComparison.OrdinalIgnoreCase))
+        {
+            return "application/xml";
+        }
```
MWO1024, line 49: is it `application/xml` or `application/ebix`?

RasmusGraabaek: ebix

RasmusGraabaek (approved):

PR #1253, opened 2024-09-24
chore: Move archived messages search tests

RasmusGraabaek:
## Checklist
- [ ] Reference to the task


source/ArchivedMessages.IntegrationTests/SearchMessagesWithAuthRestrictionTests.cs
```diff
@@ -0,0 +1,214 @@
...
+                messageWithoutRelation.Id.Value,
+                messageWithRelation.Id.Value,
+                messageWithRelation2.Id.Value,
+            ]);
+    }
+
+    /// <summary>
+    /// The following test is almost the same as the previous one, but messageWithRelation2 has a different receiver!.
+    /// </summary>
+    [Fact]
```
RasmusGraabaek, line 160: We have to discuss this

source/ArchivedMessages.IntegrationTests/SearchMessageWithoutAuthRestrictionTests.cs
```diff
@@ -0,0 +1,458 @@
...
+{
+    private readonly IArchivedMessagesClient _sut;
+    private readonly ArchivedMessagesFixture _fixture;
+
+    private readonly ActorIdentity _authenticatedActor = new(
+        ActorNumber.Create("1234512345811"),
+        Restriction.None,
+        ActorRole.MeteredDataAdministrator);
+
+    public SearchMessageWithoutAuthRestrictionTests(ArchivedMessagesFixture fixture)
```
MadsDue, line 39:
```suggestion
    public SearchMessageWithoutRestrictionTests(ArchivedMessagesFixture fixture)
```

source/ArchivedMessages.IntegrationTests/SearchMessagesWithAuthRestrictionTests.cs
```diff
@@ -0,0 +1,214 @@
...
+using Energinet.DataHub.EDI.BuildingBlocks.Domain.Authentication;
+using Energinet.DataHub.EDI.BuildingBlocks.Domain.Models;
+using FluentAssertions;
+using FluentAssertions.Execution;
+using Xunit;
+
+namespace Energinet.DataHub.EDI.ArchivedMessages.IntegrationTests;
+
+[Collection(nameof(ArchivedMessagesCollection))]
+public class SearchMessagesWithAuthRestrictionTests
```
MadsDue, line 26:
```suggestion
public class SearchMessagesWithOwnedRestrictionTests
```

source/ArchivedMessages.IntegrationTests/SearchMessagesWithAuthRestrictionTests.cs
MadsDue: I don't think we need this test, because it is not a behavior we need in EDI.

Otherwise, we should enforce that receivers should always be the same and build a test case for that.

MadsDue (approved):

PR #1257, opened 2024-09-26
refac: MasterData.IntegrationTests

lasrinnil:
## Description
Integration tests that were covering behavior for the `MasterData`module have been move into their own project.
- Upcoming work
  - Split tests that depends on both `IntegrationEvents` and `MasterData` modules
  - Create a project to share test tools between integrationprojects

image

## References
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/307
## Checklist
- [x] Reference to the task


source/EDI.sln
```diff
@@ -116,6 +116,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{2C913160
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ArchivedMessages.IntegrationTests", "ArchivedMessages.IntegrationTests\ArchivedMessages.IntegrationTests.csproj", "{E2346172-856B-44F3-8B64-AADD08BBBEA3}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{18455100-41B7-4B88-B228-DE267E460DFF}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MasterData.Integrationtests", "MasterData.Integrationtests\MasterData.Integrationtests.csproj", "{53FB8324-46B1-418F-9120-041DA95F2CF7}"
```
ebbeknudsen, line 121: Should this be `IntegrationTests` (with a big T)?

source/MasterData.Integrationtests/Fixture/Database/EdiDatabaseManager.cs
```diff
@@ -0,0 +1,87 @@
...
+                               DECLARE @sql NVARCHAR(MAX) = N'';
+
+                               SELECT @sql += 'DELETE FROM ' + QUOTENAME(TABLE_SCHEMA) + '.' + QUOTENAME(TABLE_NAME) + ' '
+                               FROM INFORMATION_SCHEMA.TABLES
+                               WHERE TABLE_TYPE = 'BASE TABLE';
+
+                               select @sql += ';'
+
+                               EXEC sp_executesql @sql;
+                               """;
```
ebbeknudsen, line 58: I'm not sure this script works if there are foreign key dependencies between tables (and the tables has data), since then tables needs to be deleted in a correct order, or the foreign key checks needs to be disabled)

source/MasterData.Integrationtests/Fixture/Database/EdiDatabaseManager.cs
```diff
@@ -0,0 +1,87 @@
...
+    public override DbContext CreateDbContext() => CreateDbContext<DbContext>();
+
+    public TDatabaseContext CreateDbContext<TDatabaseContext>()
+        where TDatabaseContext : DbContext
+    {
+        var optionsBuilder = new DbContextOptionsBuilder<TDatabaseContext>()
+            .UseSqlServer(ConnectionString, options =>
+            {
+                options.UseNodaTime();
+                options.EnableRetryOnFailure();
```
ebbeknudsen, line 40: Not sure `EnableRetryOnFailure` should be called, since we've removed it from the actual `DbContext` DI setup.

source/MasterData.Integrationtests/Fixture/MasterDataIntegrationTestCollection.cs
```diff
@@ -12,13 +12,14 @@
...
 
-using System;
+using Xunit;
 
-namespace Energinet.DataHub.EDI.IntegrationTests.Application.Actors;
+namespace Energinet.DataHub.EDI.MasterData.IntegrationTests.Fixture;
 
-internal static class SampleData
+[CollectionDefinition(nameof(MasterDataTestCollectionFixture))]
+public class MasterDataTestCollectionFixture : ICollectionFixture<MasterDataFixture>
```
ebbeknudsen, line 20:
```suggestion
public class MasterDataTestCollection : ICollectionFixture<MasterDataFixture>
```
Rename since it is a collection, not a fixture. Else the file should be renamed to match the class name.

source/MasterData.Integrationtests/Fixture/MasterDataFixture.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Energinet.DataHub.EDI.MasterData.IntegrationTests.Fixture;
+
+public class MasterDataFixture : IDisposable, IAsyncLifetime
+{
+    private bool _disposed;
+
+    public EdiDatabaseManager DatabaseManager { get; } = new();
+
+    public ServiceProvider ServiceProvider { get; private set; } = null!;
```
ebbeknudsen, line 27: What is the `ServiceProvider` on the fixture used for?

source/MasterData.Integrationtests/Fixture/MasterDataFixture.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            return;
+        }
+
+        if (disposing)
+        {
+            DatabaseManager.DeleteDatabase();
+        }
+
+        _disposed = true;
```
ebbeknudsen, line 58: If the class is marked as `sealed` we can avoid having to write this dispose pattern as far as i know.

source/MasterData.Integrationtests/Fixture/MasterDataFixture.cs
lasrinnil: For resolving class's used in tests like: 

 var sut = Services.GetService<IDataRetention>()

source/MasterData.Integrationtests/Fixture/MasterDataTestBase.cs
```diff
@@ -0,0 +1,76 @@
...
+        services.Add(ServiceDescriptor.Singleton(typeof(Logger<>), typeof(Logger<>)));
+        services.Add(ServiceDescriptor.Transient(typeof(ILogger<>), typeof(TestLogger<>)));
+
+        Services = services.BuildServiceProvider();
+    }
+
+    public void CleanupDatabase()
+    {
+        _masterDataFixture.DatabaseManager.CleanupDatabase();
+    }
```
ebbeknudsen, line 75: I'm not sure this is needed since all the test classes already has the `MasterDataFixture` and can call `fixture.DatabaseManager.CleanupDatabase()` on those instead of going through the base class.

source/MasterData.Integrationtests/Tests/RemoveOldGridAreaOwnersWhenADayHasPassedTests.cs
```diff
@@ -14,27 +14,26 @@
...
-    private readonly IMasterDataClient _masterDataClient;
+    private readonly IMasterDataClient? _masterDataClient;
 
-    public RemoveOldGridAreaOwnersWhenADayHasPassedTests(IntegrationTestFixture integrationTestFixture, ITestOutputHelper testOutputHelper)
+    public RemoveOldGridAreaOwnersWhenADayHasPassedTests(MasterDataFixture integrationTestFixture, ITestOutputHelper testOutputHelper)
         : base(integrationTestFixture, testOutputHelper)
     {
-        _masterDataClient = GetService<IMasterDataClient>();
+        SetupServiceCollection();
+        _masterDataClient = Services.GetService<IMasterDataClient>();
```
ebbeknudsen, line 36: We can use `GetRequiredService<>` instead, and then remove nullable from `_masterDataClient`

source/MasterData.Integrationtests/Tests/RemoveOldGridAreaOwnersWhenADayHasPassedTests.cs
```diff
@@ -57,16 +56,7 @@ public async Task Clean_up_grid_area_owners__has_single_owner()
...
-            .AddEnvironmentVariables()
-            .Build();
-
-        var scopedServiceCollection = new ServiceCollection();
-        scopedServiceCollection.AddMasterDataModule(config);
-        scopedServiceCollection.AddScoped<IClock>(
-            _ => new ClockStub(Instant.FromUtc(2023, 11, 3, 0, 0, 0)));
-
-        var sut = scopedServiceCollection.BuildServiceProvider().GetService<IDataRetention>()
+        var sut = Services.GetService<IDataRetention>()
```
ebbeknudsen, line 59: Can use `GetRequiredService<>` instead

source/MasterData.Integrationtests/Tests/RemoveOldGridAreaOwnersWhenADayHasPassedTests.cs
```diff
@@ -98,16 +88,7 @@ public async Task Clean_up_grid_area_owners__does_not_delete_non_expired_grid_ow
...
```
ebbeknudsen, line 91: Can use `GetRequiredService<>` instead

source/MasterData.Integrationtests/Tests/RemoveOldGridAreaOwnersWhenADayHasPassedTests.cs
```diff
@@ -152,16 +133,7 @@ public async Task Clean_up_grid_area_owners_with_multiple_owners__all_grid_areas
...
```
ebbeknudsen, line 136: Can use `GetRequiredService<>` instead

source/MasterData.Integrationtests/Tests/WhenActorIsCreatedTests.cs
```diff
@@ -12,46 +12,47 @@
...
-    public CreateActorsTests(IntegrationTestFixture integrationTestFixture, ITestOutputHelper testOutputHelper)
-        : base(integrationTestFixture, testOutputHelper)
+    public WhenActorIsCreatedTests(MasterDataFixture masterDataFixture, ITestOutputHelper testOutputHelper)
+        : base(masterDataFixture, testOutputHelper)
     {
-        _masterDataClient = GetService<IMasterDataClient>();
-        _connectionFactory = GetService<IDatabaseConnectionFactory>();
+        SetupServiceCollection();
+        _masterDataClient = Services.GetService<IMasterDataClient>();
+        _connectionFactory = Services.GetService<IDatabaseConnectionFactory>();
```
ebbeknudsen, line 37: Can use `GetRequiredService<>` instead

source/MasterData.Integrationtests/Fixture/MasterDataFixture.cs
lasrinnil: that was from the `MasterDataTestBase` the one you pointed out, I have removed now 👍

source/MasterData.Integrationtests/Tests/WhenActorIsCreatedTests.cs
```diff
@@ -62,38 +63,38 @@ public async Task Actor_is_not_created_multiple_times()
...
     {
-        using var connection = await _connectionFactory.GetConnectionAndOpenAsync(CancellationToken.None);
+        using var connection = await _connectionFactory!.GetConnectionAndOpenAsync(CancellationToken.None);
         var sql =
             $"SELECT Id, ActorNumber, ExternalId " +
             $"FROM [dbo].[Actor] " +
-            $"WHERE ExternalId = '{SampleData.ExternalId}' " +
-            $"AND ActorNumber = '{SampleData.ActorNumber}'";
+            $"WHERE ExternalId = '{ExternalId}' " +
+            $"AND ActorNumber = '{ActorNumber}'";
```
ebbeknudsen, line 97: I'm confused, does this get all actors or only specific actors? And how are we confident in our tests if we only retrieve the actors from the database that we actually expect to be created? ⚠️ .

I see that this hasn't changed in this PR, so we should probably just leave it "as-is" since this PR isn't about changing the tests.

source/MasterData.Integrationtests/Fixture/MasterDataFixture.cs
lasrinnil: like in this commit? https://github.com/Energinet-DataHub/opengeh-edi/pull/1257/commits/df94806cb9a3c33d49ee9e863c20127f3b00e31f

ebbeknudsen (approved):
LGTM

source/MasterData.Integrationtests/Tests/WhenActorIsCreatedTests.cs
lasrinnil: I believe that the `GetAllActors()` is to ensure that only the expected `Actors` are present as pr. the behavior of the test.

PR #1259, opened 2024-09-27
refac: IntegrationEvents.IntegrationTests

lasrinnil:
To use the same approach for the new `IntegrationTests` projects as in this PR: https://github.com/Energinet-DataHub/opengeh-edi/pull/1257  this PR will conform `IntegrationEvents.IntegrationTests`

## References
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/308
## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/11167098681
- [x] Reference to the task


source/IntegrationEvents.IntegrationTests/Fixture/IntegrationEventsFixture.cs
```diff
@@ -30,28 +28,7 @@ public async Task InitializeAsync()
 
     public async Task DisposeAsync()
     {
-        Dispose();
```
ebbeknudsen, line 33: Isn't this missing `DatabaseManager.DeleteDatabase()` instead of `DatabaseManager.CleanupDatabase()`

source/IntegrationEvents.IntegrationTests/Fixture/IntegrationEventsTestBase.cs
```diff
@@ -36,7 +36,7 @@ public IntegrationEventsTestBase(IntegrationEventsFixture integrationEventsFixtu
         _integrationEventsFixture = integrationEventsFixture;
         _testOutputHelper = testOutputHelper;
 
-        CleanupDatabase();
```
ebbeknudsen, line 39: Maybe i missed this in the previous PR, but I don't think it is "clean" to do actual work in the constructor. I think it would be more correct to implement `IAsyncLIfetime` and perform actual work in the `InitializeAsync()` and `DisposeAsync()` lifecycle methods.

Since this is more a matter of style/opinion i will let you decide 👍.

lasrinnil: Agreed. I have updated the PR to reflect your suggestion 👍

source/IntegrationEvents.IntegrationTests/Fixture/IntegrationEventsFixture.cs
lasrinnil: It is. I am making a new PR to correct just this change in the `MasterData.IntegrationTests`
https://github.com/Energinet-DataHub/opengeh-edi/pull/1264

MadsDue (approved):

PR #1260, opened 2024-09-27
feat: pagination and sorting of archived messages

MadsDue:
## Description
We want to avoid loading all archived messages into memory when searching. This pull request introduces pagination for archived messages, allowing users to navigate forwards and backwards through the dataset without loading more data than necessary into memory. It also supports sorting, enabling users to choose a specific field to sort the dataset by.
## References

BFF pr
- https://github.com/Energinet-DataHub/greenforce-frontend/pull/3596

## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/11143947880
- [x] Reference to the task => https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/312


source/ArchivedMessages.Infrastructure/QueryBuilder.cs
```diff
@@ -95,15 +95,47 @@ public QueryInput BuildFrom(GetMessagesQuery request)
...
+                : directionToSortBy == DirectionToSortBy.Descending ? ">" : "<";
+        return isForward
+            ? $"""
+                  ({fieldToSortBy.Identifier} = '{cursor.SortedFieldValue}' AND (RecordId < {cursor.RecordId} OR {cursor.RecordId} = 0)) 
+                  OR ({fieldToSortBy.Identifier} {sortingDirection} '{cursor.SortedFieldValue}')
+              """
+            : $"""
+                   ({fieldToSortBy.Identifier} = '{cursor.SortedFieldValue}' AND (RecordId > {cursor.RecordId} OR {cursor.RecordId} = 0)) 
+                   OR ({fieldToSortBy.Identifier} {sortingDirection} '{cursor.SortedFieldValue}') 
+               """;
```
MWO1024, line 120: Should we use `<=` and `>=` in case the column to sort by isn't unique?

source/ArchivedMessages.Infrastructure/QueryBuilder.cs
```diff
@@ -95,15 +95,47 @@ public QueryInput BuildFrom(GetMessagesQuery request)
...
+        // Toggle the sort direction if navigating backwards, because sql use top to limit the result
+        var sortDirection = navigatingForward ? directionToSortBy : directionToSortBy == DirectionToSortBy.Ascending ? DirectionToSortBy.Descending : DirectionToSortBy.Ascending;
+        return $" ORDER BY {fieldToSortBy.Identifier} {sortDirection.Identifier}, RecordId {pagingDirection}";
+    }
+
+    private string BuildStatement(GetMessagesQuery query)
+    {
+        var whereClause = " WHERE ";
+        whereClause += _statement.Count > 0 ? $"{string.Join(" AND ", _statement)} AND " : string.Empty;
+        whereClause += WherePaginationPosition(query.Pagination.SortBy, query.Pagination.DirectionToSortBy, query.Pagination.Cursor, query.Pagination.NavigationForward);
```
MWO1024, line 135: I wonder if readability improves if we use """ here too.

source/ArchivedMessages.Infrastructure/QueryBuilder.cs
```diff
@@ -30,55 +30,55 @@ public QueryBuilder(ActorIdentity actorIdentity)
         _actorIdentity = actorIdentity;
     }
 
-    public QueryInput BuildFrom(GetMessagesQuery request)
+    public QueryInput BuildFrom(GetMessagesQuery query)
```
MWO1024, line 33: Comment related to file in general: I wonder if this would benefit from some example queries in comments for each helper method, like we did for requests in wholesale?

source/ArchivedMessages.Interfaces/SortByDirection.cs
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Energinet.DataHub.EDI.ArchivedMessages.Interfaces;
+
+public record DirectionToSortBy(string Identifier)
```
MWO1024, line 17: Maybe `readonly record struct` is the way to go here. We are only wrapping a string and it is not something we have to store. Perhaps a plain old `readonly struct` is sufficient?

source/ArchivedMessages.Infrastructure/QueryBuilder.cs
MadsDue: if the sort column isn't unique the first line should handle it:
{**fieldToSortBy.Identifier} = '{cursor.SortedFieldValue}'** AND (**RecordId < {cursor.RecordId}** OR {cursor.RecordId} = 0))

It takes the next RecordId when the sort field value is the same :)

source/ArchivedMessages.Infrastructure/QueryBuilder.cs
MadsDue: I am not sure, because of the if statement in the middle :)

source/ArchivedMessages.Infrastructure/QueryBuilder.cs
MadsDue: I have added comments to this. I hope that the test cases are showcasing the examples :)

source/ArchivedMessages.Interfaces/SortByDirection.cs
MadsDue: Good catch :)

source/ArchivedMessages.Infrastructure/QueryBuilder.cs
MWO1024: Oh right. I'm apparently very bad at reading.

source/ArchivedMessages.IntegrationTests/SearchMessagesWithoutRestrictionTests.cs
```diff
@@ -449,10 +478,436 @@ public async Task Given_FourArchivedMessagesWithRelations_When_IncludingRelatedM
...
+                    "1234512345123",
+                    "1234512345128",
+                    DocumentType.RejectRequestWholesaleSettlement.Name),
+                new(
+                    CreatedAt("2023-04-06T22:00:00Z"),
+                    Guid.NewGuid().ToString(),
+                    "1234512345122",
+                    "1234512345127",
+                    DocumentType.NotifyAggregatedMeasureData.Name),
+            };
```
MWO1024, line 737: Missing duplicate value for sender and receiver

source/ArchivedMessages.IntegrationTests/SearchMessagesWithoutRestrictionTests.cs
```diff
@@ -449,10 +478,436 @@ public async Task Given_FourArchivedMessagesWithRelations_When_IncludingRelatedM
...
+            orderedMessages = sortedDirection.Identifier == DirectionToSortBy.Ascending.Identifier
+                ? messages.OrderBy(x => x.MessageId)
+                : messages.OrderByDescending(x => x.MessageId);
+        }
+
+        if (sortedBy.Identifier == FieldToSortBy.CreatedAt.Identifier)
+        {
+            orderedMessages = sortedDirection.Identifier == DirectionToSortBy.Ascending.Identifier
+                ? messages.OrderBy(x => x.CreatedAt)
+                : messages.OrderByDescending(x => x.CreatedAt);
```
MWO1024, line 879: One could argue we should add the `ThenByDesc` here too, but in practise, what are the chances that two rows will have the exact same `CreatedAt`-time?
I'm tempted to argue we should play it safe and add it here, and update the two tests to include duplicated `CreatedAt` cases as well.

MWO1024 (dismissed):

source/ArchivedMessages.IntegrationTests/SearchMessagesWithoutRestrictionTests.cs
MadsDue: I agree with you :)

MWO1024 (approved):

MWO1024 (dismissed):

MWO1024 (dismissed):

MWO1024 (approved):

PR #1272, opened 2024-10-02
fix: no auth actor for retention jobs

MadsDue:
## Description
FIx: System.InvalidOperationException: Current identity is not set	
in
Energinet.DataHub.EDI.BuildingBlocks.Domain.Authentication.AuthenticatedActor.get_CurrentActorIdentity.

when retention jobs is running. 

## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/11176690186
- [ ] Reference to the task


source/B2BApi/Security/AuditUserContext.cs
```diff
@@ -24,17 +24,21 @@ public AuditUser? CurrentUser
     {
         get
         {
-            var currentActor = authenticatedActor.CurrentActorIdentity;
+            if (!authenticatedActor.TryGetCurrentActorIdentity(out var currentActor))
+            {
+                // In some cases, we don't have an authenticated actor, e.g. when we are running in a background service like retention jobs.
+                return null;
+            }
```
ebbeknudsen, line 31: Would it make sense to move this into the `AuditLogger` instead? And then maybe the audit log method call could include if a actor is required or not.

Reason: This could potentially mean that an error happened in getting the current actor (for an audit log where that is required), and then we log the auditlog anyway, as if nothing went wrong.

RasmusGraabaek: HostFactory.cs

PR #1279, opened 2024-10-07
feat: metric on number of outgoing messages

RasmusGraabaek:
## Checklist
- [ ] Reference to the task


source/Tests/Application/OutgoingMessages/MetricNameGenerationTests.cs
```diff
@@ -0,0 +1,82 @@
...
+                    names.Add(MetricNameMapper.MessageGenerationMetricName(
+                        documentType,
+                        documentFormat,
+                        true));
+                }
+            }
+        }
+
+        return names;
+    }
```
MWO1024, line 81: `TheoryData`?

MWO1024 (approved):

PR #1292, opened 2024-10-11
feat: run retention jobs async

MadsDue:
## Description
Run out retention jobs async

## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/11291056600
- [ ] Reference to the task


source/B2BApi/DataRetention/ExecuteDataRetentionsWhenADayHasPassed.cs
```diff
@@ -24,35 +24,65 @@ public class ExecuteDataRetentionsWhenADayHasPassed : INotificationHandler<ADayH
 {
     private readonly IReadOnlyCollection<IDataRetention> _dataRetentions;
     private readonly ILogger<ExecuteDataRetentionsWhenADayHasPassed> _logger;
+    private readonly int _jobsExecutionTimeLimitInSeconds;
 
     public ExecuteDataRetentionsWhenADayHasPassed(
         IEnumerable<IDataRetention> dataRetentions,
-        ILogger<ExecuteDataRetentionsWhenADayHasPassed> logger)
+        ILogger<ExecuteDataRetentionsWhenADayHasPassed> logger,
+        int executionTimeLimitInSeconds = 25 * 60)
```
ebbeknudsen, line 32: I'm unsure, does default values in constructors work with dependency injection?

source/B2BApi/DataRetention/ExecuteDataRetentionsWhenADayHasPassed.cs
```diff
@@ -24,35 +24,65 @@ public class ExecuteDataRetentionsWhenADayHasPassed : INotificationHandler<ADayH
...
+            var tasks = _dataRetentions.Select(async dataCleaner =>
             {
-                TimeSpan.FromSeconds(10),
-                TimeSpan.FromSeconds(15),
-                TimeSpan.FromSeconds(30),
+                var task = executionPolicy.ExecuteAsync(
+                    () => dataCleaner.CleanupAsync(linkedCts.Token));
+                taskMap[task] = dataCleaner;
+                await task.ConfigureAwait(false);
             });
```
ebbeknudsen, line 63: I think using "await" here stops the `.Select` loop, so I'm not sure the next datacleaner is started before the previous one has run to completion. Correct me if I'm wrong, but i think it should be something like this instead:

```csharp
var tasks = _dataRetentions.Select(dataCleaner =>
{
    var task = executionPolicy.ExecuteAsync(
        () => dataCleaner.CleanupAsync(linkedCts.Token));
    taskMap[task] = dataCleaner;
    return task;
});
```

Or maybe to make it more clear what actually happens:
```csharp
foreach (var dataCleaner in _dataRetentions)
{
    var task = executionPolicy.ExecuteAsync(
        () => dataCleaner.CleanupAsync(linkedCts.Token));

    taskMap[task] = dataCleaner;
}

await Task.WhenAll(taskMap.Select(kv => kv.Key))
    .ConfigureAwait(false);
``` 

This makes it clear that the all the tasks are started, and the `await` only happens at the `Task.WhenAll()`.

source/B2BApi/DataRetention/ExecuteDataRetentionsWhenADayHasPassed.cs
```diff
@@ -24,35 +24,65 @@ public class ExecuteDataRetentionsWhenADayHasPassed : INotificationHandler<ADayH
...
 
     public async Task Handle(ADayHasPassed notification, CancellationToken cancellationToken)
     {
-        var executionPolicy = Policy
-            .Handle<Exception>()
-            .WaitAndRetryAsync(new[]
+        var taskMap = new Dictionary<Task, IDataRetention>();
+        try
+        {
+            // Cancels all retentions after provided minutes
```
ebbeknudsen, line 44:
```suggestion
            // Cancels all retentions after provided seconds
```

source/B2BApi/DataRetention/ExecuteDataRetentionsWhenADayHasPassed.cs
```diff
@@ -24,35 +24,67 @@ public class ExecuteDataRetentionsWhenADayHasPassed : INotificationHandler<ADayH
...
-                _logger?.Log(LogLevel.Error, result.FinalException, "Type {DataCleaner} failed to clean up", dataCleaner.GetType().FullName);
+                var task = executionPolicy.ExecuteAsync(
+                    () => dataCleaner.CleanupAsync(linkedCts.Token));
+
+                taskMap[task] = dataCleaner;
             }
+
+            var tasks = taskMap.Keys.ToList();
+
+            await Task.WhenAll(tasks).ConfigureAwait(false);
```
ebbeknudsen, line 67: Sadly Entity Framework DbContext's doesn't support parallel operations: https://learn.microsoft.com/en-us/ef/core/miscellaneous/async so we have to change the code to use a different `DbContext` (Dependency Injection scope) for each retention job.

lasrinnil (approved):

PR #1295, opened 2024-10-16
feat: initial code move for requesting timeseries

lasrinnil:
## Description
**None of the added code with this PR is active**

This will be the initial move of the code from the Wholesale repo that handles the validation of request and retrieving for timeseries.

This move also includes `CalculationResults`. But the enabling the tests for parts of that module is made in an upcoming pr.

## References
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/338
## Checklist
- [x] Reference to the task


source/Edi.UnitTests/Edi.UnitTests.csproj
```diff
@@ -0,0 +1,42 @@
...
+  <ItemGroup>
+    <PackageReference Include="Energinet.DataHub.Core.App.Common" Version="13.2.0" />
+    <PackageReference Include="Energinet.DataHub.Core.TestCommon" Version="7.0.0" />
+    <PackageReference Include="FluentAssertions" Version="6.12.1" />
+    <PackageReference Include="FluentAssertions.Analyzers" Version="0.33.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="8.0.8" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="8.0.0" />
```
MWO1024, line 17: EntityFramework does not smell of unit tests, but there is probably some reference in the code under test, that requires this regardless. What would happen in this reference was to disappear? Could we locate the sinner?

MWO1024 (approved):
Let us get the code moved. Then we can delete, move, refactor, and all the other fun things later.

djorgensendk (approved):

PR #1305, opened 2024-10-24
feat: impl metered data for measurement point ebix parser

MadsDue:
## Checklist
- [ ] Reference to the task


source/IncomingMessages.Infrastructure/MessageParsers/BaseParsers/Ebix/MessageHeaderExtractor.cs
```diff
@@ -0,0 +1,64 @@
...
+    private const string RecipientEnergyParty = "RecipientEnergyParty";
+    private const string EnergyBusinessProcess = "EnergyBusinessProcess";
+    private const string EnergyBusinessProcessRole = "EnergyBusinessProcessRole";
+    private const string EnergyIndustryClassification = "EnergyIndustryClassification";
+
+    public static MessageHeader Extract(
+        XDocument document,
+        XNamespace ns)
+    {
+        var headerElement = document.Descendants(ns + HeaderElementName).FirstOrDefault();
```
MWO1024, line 36: Is it ok if there is more than one header element? Not sure if we should use `.SingleOrDefault` here or not. Might be a good idea if the schema says so (i.e. only one header element allowed), but at the same time, we need to make sure the schema is validated before we fail here then.

source/IncomingMessages.Infrastructure/MessageParsers/MeteredDateForMeasurementPointParsers/Ebix/MeteredDataForMeasurementPointSeriesExtractor.cs
```diff
@@ -0,0 +1,77 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Xml.Linq;
+using Energinet.DataHub.EDI.IncomingMessages.Domain;
+
+namespace Energinet.DataHub.EDI.IncomingMessages.Infrastructure.MessageParsers.MeteredDateForMeasurementPointParsers.Ebix;
+
+public class MeteredDataForMeasurementPointSeriesExtractor
```
MWO1024, line 20: Should we make the class `static` if all its methods are static? Just to avoid confusion (no one should make instances of this class)

source/Tests/CimMessageAdapter/Messages/MeteredDateForMeasurementPointEbixMessageParserTests/MessageParserTests.cs
```diff
@@ -0,0 +1,149 @@
...
+            [
+                DocumentFormat.Ebix, CreateBaseEbixMessage("BadVersionMeteredDateForMeasurementPoint.xml"),
+                nameof(InvalidBusinessReasonOrVersion),
+            ],
+            [
+                DocumentFormat.Ebix, CreateBaseEbixMessage("InvalidMeteredDateForMeasurementPoint.xml"),
+                nameof(InvalidMessageStructure),
+            ],
+        ];
+    }
```
MWO1024, line 66: `TheoryData`?

source/Tests/CimMessageAdapter/Messages/MeteredDateForMeasurementPointEbixMessageParserTests/MessageParserTests.cs
```diff
@@ -0,0 +1,149 @@
...
+                .Match(resolution => resolution == "PT15M" || resolution == "PT1H");
+            series.StartDateTime.Should()
+                .Match(startDate => startDate == "2024-06-28T22:00:00Z" || startDate == "2024-06-29T22:00:00Z");
+            series.EndDateTime.Should()
+                .Match(endDate => endDate == "2024-06-29T22:00:00Z" || endDate == "2024-06-30T22:00:00Z");
+            series.ProductNumber.Should().Be("8716867000030");
+            series.ProductUnitType.Should().Be("KWH");
+            series.MeteringPointType.Should().Be("E18");
+            series.MeteringPointLocationId.Should().Be("571313000000002000");
+            series.GridArea.Should().BeNull();
```
MWO1024, line 108: It would be nice to pair up all the choice values, but I'm not sure it is worth our time. The content of the files is static, so the chance we manage to create a parser where the values end up mixed is probably zero.

source/IncomingMessages.Infrastructure/MessageParsers/BaseParsers/Ebix/MessageHeaderExtractor.cs
MadsDue: Good catch, the schema ensures that there is only one of these.

source/IncomingMessages.Infrastructure/MessageParsers/MeteredDateForMeasurementPointParsers/Ebix/MeteredDataForMeasurementPointSeriesExtractor.cs
```diff
@@ -0,0 +1,77 @@
...
+    private const string UnitType = "UnitType";
+    private const string DetailMeasurementMeteringPointCharacteristic = "DetailMeasurementMeteringPointCharacteristic";
+    private const string TypeOfMeteringPoint = "TypeOfMeteringPoint";
+    private const string MeteringPointDomainLocation = "MeteringPointDomainLocation";
+    private const string Position = "Position";
+    private const string EnergyQuantity = "EnergyQuantity";
+    private const string QuantityQuality = "QuantityQuality";
+    private const string IntervalEnergyObservation = "IntervalEnergyObservation";
+
+    protected internal static IEnumerable<MeteredDataForMeasurementPointSeries> ParseSeries(
```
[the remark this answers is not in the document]
MadsDue: If we change it to a read-only collection then we can't use yield return

source/IncomingMessages.Infrastructure/MessageParsers/MeteredDateForMeasurementPointParsers/Ebix/MeteredDataForMeasurementPointSeriesExtractor.cs
MadsDue: good catch

source/Tests/CimMessageAdapter/Messages/MeteredDateForMeasurementPointEbixMessageParserTests/MessageParserTests.cs
MadsDue: Yes

source/Tests/CimMessageAdapter/Messages/MeteredDateForMeasurementPointEbixMessageParserTests/MessageParserTests.cs
MadsDue: it was just an attempt to keep it simple without wasting time. If we need more specific cases, I imagine we would create more specific ones, but I don't see the need yet.

MWO1024 (approved):

PR #1306, opened 2024-10-24
feat: Add RequestWholesaleServices orchestration

ebbeknudsen:
## Description

- Add RequestWholesaleServicesOrchestration
    - Can only enqueue "notify" messages, reject messages will be implemented later
- Use existing "Wholesale" calculation result / EDI modules, while trying to refactor as little as possible
    - Refactor and deletion of "Wholesale" code will happen in a later PR
- Add B2BApi app tests of `RequestWholesaleServicesOrchestration`

## Checklist
- [ ] Reference to the task => https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/339


source/B2BApi.AppTests/Fixtures/B2BApiAppFixture.cs
```diff
@@ -52,13 +55,14 @@ public class B2BApiAppFixture : IAsyncLifetime
...
+    private readonly DeltaTableOptions _calculationResultsDatabricksOptions;
+
     public B2BApiAppFixture()
     {
-        var constructorStopwatch = new Stopwatch();
-        var stopwatch = new Stopwatch();
-
-        constructorStopwatch.Start();
-        stopwatch.Start();
+        var constructorStopwatch = Stopwatch.StartNew();
```
lasrinnil, line 64: Are these metrics still used?

lasrinnil: If no one is actively using them, they are slowing down test execution I would pressume

source/B2BApi.AppTests/Fixtures/B2BApiAppFixture.cs
```diff
@@ -95,11 +99,23 @@ public B2BApiAppFixture()
...
         LogStopwatch(stopwatch, nameof(HostConfigurationBuilder));
 
-        DatabricksSchemaManager = new DatabricksSchemaManager(
+        EdiDatabricksSchemaManager = new DatabricksSchemaManager(
+            new HttpClientFactory(),
+            IntegrationTestConfiguration.DatabricksSettings,
+            "B2BApi_tests_edi");
+        LogStopwatch(stopwatch, nameof(EdiDatabricksSchemaManager));
+
+        CalculationResultsDatabricksSchemaManager = new DatabricksSchemaManager(
```
lasrinnil, line 108: I would have the highest confidence if we only had one configuration to Databricks and the schemas were only stated once when we want to run this in production

source/B2BApi.AppTests/Fixtures/Extensions/HttpRequestExtensions.cs
```diff
@@ -0,0 +1,131 @@
...
+    private static async Task<HttpRequestMessage> CreateHttpRequestAsync(
+        B2BApiAppFixture fixture,
+        Actor actor,
+        HttpMethod method,
+        string url,
+        HttpContent? content = null)
+    {
+        // The actor must exist in the database
+        var externalId = Guid.NewGuid().ToString();
+        await fixture.DatabaseManager.AddActorAsync(actor.ActorNumber, externalId);
```
lasrinnil, line 113: If we run a test for the same `ActorNumber` would an exception not be thrown? I would expect a constraint on `ActorNumber` and `ExternalId` for a `Actor`

source/B2BApi.AppTests/Functions/RequestWholesaleServices/RequestWholesaleServicesTests.cs
```diff
@@ -0,0 +1,212 @@
...
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Extensions.Options;
+using NodaTime;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Energinet.DataHub.EDI.B2BApi.AppTests.Functions.RequestWholesaleServices;
+
+[Collection(nameof(B2BApiAppCollectionFixture))]
+public class RequestWholesaleServicesTests : IAsyncLifetime
```
lasrinnil, line 39: I would suggest that the test in this class is moved to `IncomingMessageReceiverTests` to better reflect the pathing. We dont really have a Function named `RequestWholesaleServices`

lasrinnil: wow - sorry we now HAVE a function named `RequestWholesaleServices`

lasrinnil: Perhaps either the path or testclass should be named something with `RequestWholesaleServicesOrchestration`

source/B2BApi.AppTests/TestData/CalculationResults/AmountsPerChargeViewSchemaDefinition.cs
```diff
@@ -0,0 +1,47 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Energinet.DataHub.Wholesale.CalculationResults.Infrastructure.SqlStatements.DeltaTableConstants;
+
+namespace Energinet.DataHub.EDI.B2BApi.AppTests.TestData.CalculationResults;
+
+public class AmountsPerChargeViewSchemaDefinition
```
lasrinnil, line 19: Can we just not have a reference to these files? There is already a dependency from B2BApiTests -> IntegrationTests?

source/B2BApi.AppTests/Fixtures/B2BApiAppFixture.cs
ebbeknudsen: I'm not sure how it will work since we use 2 different methods of retreiving the calculation results (one method for when a calculation is completed, another method for when handling energy/wholesale requests), thus I have the highest confidence that the code works when we don't use the same schema manager, so these tests doesn't interfere with eachother. When we align/refactor the code i think it makes sense to use the same schema manager.

source/B2BApi/Functions/RequestWholesaleServices/Activities/EnqueueMessagesForRequestWholesaleServicesActivity.cs
```diff
@@ -0,0 +1,106 @@
...
+    RequestWholesaleServicesQueryHandler requestWholesaleServicesQueryHandler,
+    IOutgoingMessagesClient outgoingMessagesClient,
+    IUnitOfWork unitOfWork)
+{
+    private readonly RequestWholesaleServicesQueryHandler _requestWholesaleServicesQueryHandler = requestWholesaleServicesQueryHandler;
+    private readonly IOutgoingMessagesClient _outgoingMessagesClient = outgoingMessagesClient;
+    private readonly IUnitOfWork _unitOfWork = unitOfWork;
+
+    /// <summary>
+    /// Start an ValidateWholesaleServicesRequestActivity activity.
```
lasrinnil, line 39:
```suggestion
    /// Start an EnqueueMessagesForRequestWholesaleServices activity.
```

source/B2BApi/Functions/RequestWholesaleServices/Activities/EnqueueMessagesForRequestWholesaleServicesActivity.cs
```diff
@@ -0,0 +1,106 @@
...
+                // TODO: Enqueue rejected message
+                // var rejectedWholesaleServicesMessage = RejectedWholesaleServiceMessageDtoFactory.Create();
+                // var messageId = await _outgoingMessagesClient.EnqueueAsync(rejectedWholesaleServicesMessage, cancellationToken)
+                //     .ConfigureAwait(false);
+                rejectedMessageIds.Add(Guid.Empty);
+            }
+        }
+
+        // TODO: Should this save per transaction (as it works today) or is it better to save once when all messages are enqueued?
+        await _unitOfWork.CommitTransactionAsync(cancellationToken)
```
lasrinnil, line 91: Better to commit after alle messages are enqueued, this way we wont need to idempotent in case the activity fails.

lasrinnil, line 91: Perhaps we should ask the business if it is ok to not deliver all the results at the same time?

source/B2BApi/Functions/RequestWholesaleServices/Activities/EnqueueMessagesForRequestWholesaleServicesActivity.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        var request = WholesaleServicesRequestFactory.CreateWholesaleServicesRequest(transaction);
+        var calculationResultsAsyncEnumerable = _requestWholesaleServicesQueryHandler.GetAsync(
+            request,
+            transaction.BusinessTransactionId.Value,
+            cancellationToken);
+
+        List<Guid> acceptedMessagesIds = [];
+        List<Guid> rejectedMessageIds = [];
+        await foreach (var calculationResult in calculationResultsAsyncEnumerable)
```
lasrinnil, line 65: ☎️

source/B2BApi/Functions/RequestWholesaleServices/Activities/GetWholesaleServicesRequestTransactionsActivity.cs
```diff
@@ -0,0 +1,68 @@
...
+using Energinet.DataHub.Wholesale.Edi.Models;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.DurableTask;
+
+namespace Energinet.DataHub.EDI.B2BApi.Functions.RequestWholesaleServices.Activities;
+
+public class GetWholesaleServicesRequestTransactionsActivity
+{
+    /// <summary>
+    /// Start an ValidateWholesaleServicesRequestActivity activity.
```
lasrinnil, line 26:
```suggestion
    /// Start an EnqueueMessagesForRequestWholesaleServices activity.
```

source/B2BApi/Functions/RequestWholesaleServices/RequestWholesaleServicesOrchestration.cs
```diff
@@ -0,0 +1,80 @@
...
+            return "Error: No input given.";
+
+        // Split request in multiple transactions
+        var transactions = await GetWholesaleServicesRequestTransactionsActivity
+            .StartActivityAsync(
+                input,
+                context,
+                null);
+
+        // TODO: Maybe use sub orchestrations instead of fan-out?
```
lasrinnil, line 41: I would suggest we have the default approach to parallelization be fan-out. it is a used when we enqueue messages after calculations are completed.

ebbeknudsen: I'm using fan-out as well (it's simpler), since this will be refactored when we refactor to use the process manager.

source/B2BApi.AppTests/TestData/CalculationResults/AmountsPerChargeViewSchemaDefinition.cs
ebbeknudsen: The `AmountsPerChargeViewSchemaDefinition` is a little different between them (for now), which is why it is copied.

source/B2BApi.AppTests/Fixtures/Extensions/HttpRequestExtensions.cs
ebbeknudsen: The Actor is only unique if both `ExternalId` and `ActorNumber` is the same, which it never is since `externalId` is a new guid each time this is called.

lasrinnil (approved):

lasrinnil (approved):

PR #1311, opened 2024-10-25
test: Handle AggregatedMeasuredData only in EDI

lasrinnil:
## Checklist
- [ ] Reference to the task

### DIFF
--- a/source/B2BApi.AppTests/Fixtures/B2BApiAppFixture.cs
+++ b/source/B2BApi.AppTests/Fixtures/B2BApiAppFixture.cs
@@ -417,7 +417,7 @@ public class B2BApiAppFixture : IAsyncLifetime
 
         appHostSettings.ProcessEnvironmentVariables.Add(
             "FeatureManagement__RequestStaysInEdi",
-            false.ToString().ToLower());
+            true.ToString().ToLower());
 
         appHostSettings.ProcessEnvironmentVariables.Add(
             $"RevisionLogOptions__{nameof(RevisionLogOptions.ApiAddress)}",
@@ -425,7 +425,7 @@ public class B2BApiAppFixture : IAsyncLifetime
 
         appHostSettings.ProcessEnvironmentVariables.Add(
             $"AzureWebJobs.TenSecondsHasPassed.Disabled",
-            "true");
+            "false");
         appHostSettings.ProcessEnvironmentVariables.Add(
             $"AzureWebJobs.ADayHasPassed.Disabled",
             "true");
--- a/source/B2BApi.AppTests/IncomingMessages/IncomingMessageReceiverTests.cs
+++ b/source/B2BApi.AppTests/IncomingMessages/IncomingMessageReceiverTests.cs
@@ -84,6 +84,61 @@ public class IncomingMessageReceiverTests : IAsyncLifetime
     }
 
     [Fact]
+    public async Task Given_PersistedActor_When_CallingIncomingMessagesWithValidDocumentAndBearerToken_Then_ResponseShouldBeAccepted_AndRequestIsPeekable()
+    {
+        using var request = await CreateHttpRequest(
+            "TestData/Messages/json/RequestAggregatedMeasureData.json",
+            IncomingDocumentType.RequestAggregatedMeasureData.Name,
+            "application/json");
+
+        // Act
+        using var actualResponse = await Fixture.AppHostManager.HttpClient.SendAsync(request);
+
+        // Assert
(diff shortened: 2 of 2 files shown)


source/B2BApi.AppTests/Fixtures/B2BApiAppFixture.cs
```diff
@@ -429,7 +429,7 @@ private FunctionAppHostSettings CreateAppHostSettings(string csprojName, ref int
 
         appHostSettings.ProcessEnvironmentVariables.Add(
             $"AzureWebJobs.TenSecondsHasPassed.Disabled",
-            "true");
+            "false");
```
ebbeknudsen, line 432: This is meant to be disabled because it muddies the tests. It can be manually triggered in the test by running something like this: `await Fixture.AppHostManager.TriggerFunctionAsync(nameof(OutboxPublisher));`

source/B2BApi.AppTests/IncomingMessages/IncomingMessageReceiverTests.cs
```diff
@@ -83,6 +84,54 @@ public async Task Given_PersistedActor_When_CallingIncomingMessagesWithValidDocu
...
+            if (response.StatusCode == HttpStatusCode.OK)
+            {
+                // Assert
+                response.Should().NotBeNull();
+                response.StatusCode.Should().Be(HttpStatusCode.OK);
+                (await response.Content.ReadAsStringAsync()).Should().Contain("RejectRequestAggregatedMeasureData_MarketDocument");
+            }
+
+            await Task.Delay(500);
+        }
```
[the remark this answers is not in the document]
lasrinnil: I will look into it (y)

source/B2BApi.AppTests/IncomingMessages/IncomingMessageReceiverTests.cs
```diff
@@ -83,6 +84,54 @@ public async Task Given_PersistedActor_When_CallingIncomingMessagesWithValidDocu
...
+                "application/json");
+            peekRequest.Headers.Authorization = new AuthenticationHeaderValue("bearer", b2bToken);
+            using var response = await Fixture.AppHostManager.HttpClient.SendAsync(peekRequest);
+
+            if (response.StatusCode == HttpStatusCode.OK)
+            {
+                // Assert
+                response.Should().NotBeNull();
+                response.StatusCode.Should().Be(HttpStatusCode.OK);
+                (await response.Content.ReadAsStringAsync()).Should().Contain("RejectRequestAggregatedMeasureData_MarketDocument");
```
ebbeknudsen, line 128: If the statuscode is not `HttpStatusCode.OK` this will never output the response content (since the tests stops at the statuscode assertion, thus making it hard to find the cause of the test failing. This could be fixed with an assertion scope around the statuscode and content.

source/B2BApi.AppTests/IncomingMessages/IncomingMessageReceiverTests.cs
```diff
@@ -83,6 +84,54 @@ public async Task Given_PersistedActor_When_CallingIncomingMessagesWithValidDocu
...
+        var actorRole = ActorRole.EnergySupplier;
+        var b2bToken = new JwtBuilder()
+            .WithRole(ClaimsMap.RoleFrom(actorRole).Value)
+            .WithClaim(ClaimsMap.ActorId, externalId)
+            .CreateToken();
+
+        using var request = await CreateHttpRequest(
+            "TestData/Messages/json/RequestAggregatedMeasureData.json",
+            IncomingDocumentType.RequestAggregatedMeasureData.Name,
+            "application/json");
```
ebbeknudsen, line 104: I assume the actor number `5790000392551` and role should match the one on the `RequestAggregatedMeasureData.json`, however it is not very clear in the way this code is written.

source/B2BApi.AppTests/IncomingMessages/IncomingMessageReceiverTests.cs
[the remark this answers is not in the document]
lasrinnil: I dont really feel its a fit in this case. If we could go over it together it would be nice 👍

source/B2BApi.AppTests/IncomingMessages/IncomingMessageReceiverTests.cs
lasrinnil: Agreed. The actor in case have been removed from the test and is now setup for the test class.

PR #1317, opened 2024-10-29
feat: ProcessManager API and Client

dstenroejl:
## Description

Walking skeleton for ProcessManager API and Client.

Deployed to `dev_002` to verify we didn't break anything: https://github.com/Energinet-DataHub/dh3-environments/actions/runs/11668859125

## References

https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/331

## Checklist
- [ ] Reference to the task


source/ProcessManager.Client.Tests/Fixtures/ScenarioAppFixturesConfiguration.cs
```diff
@@ -0,0 +1,51 @@
...
+    private static readonly Lazy<ScenarioAppFixturesConfiguration> _instance =
+        new(() => new ScenarioAppFixturesConfiguration());
+
+    private ScenarioAppFixturesConfiguration()
+    {
+        DatabaseManager = new ProcessManagerDatabaseManager("ClientsTest");
+        TaskHubName = "ClientsTest01";
+        OrchestrationsAppPort = 8101;
+        ProcessManagerAppPort = 8102;
+    }
```
ebbeknudsen, line 40: Is there maybe a potential issue in that this `DatabaseManager` is a static singleton, so the lifecycle isn't managed by the fixture's anymore? (which means it will be reused between fixtures)

source/ProcessManager.Client.Tests/Integration/MonitorCalculationScenarioUsingApi.cs
```diff
@@ -0,0 +1,130 @@
...
+using Xunit.Abstractions;
+
+namespace Energinet.DataHub.ProcessManager.Client.Tests.Integration;
+
+/// <summary>
+/// Test case where we verify the Process Manager clients can be used to start a
+/// calculation orchestration and monitor its status during its lifetime.
+/// </summary>
+[Collection(nameof(ProcessManagerClientCollection))]
+public class MonitorCalculationScenarioUsingApi : IAsyncLifetime
```
ebbeknudsen, line 33:
```suggestion
public class MonitorCalculationScenarioUsingApiTests : IAsyncLifetime
```

source/ProcessManager.Client/Processes/BRS_023_027/V1/INotifyAggregatedMeasureDataClientV1.cs
```diff
@@ -0,0 +1,54 @@
...
+
+/// <summary>
+/// Client for using the BRS-023/BRS_027 Process Manager API.
+/// </summary>
+public interface INotifyAggregatedMeasureDataClientV1
+{
+    /// <summary>
+    /// Schedule a BRS-023 or BRS-027 calculation orchestration instance and return its id.
+    /// </summary>
+    public Task<Guid> ScheduleNewCalculationOrchestationInstanceAsync(
```
ebbeknudsen, line 29:
```suggestion
    public Task<Guid> ScheduleNewCalculationAsync(
```

Since this is the _public specific_ client for calculations, I think we should omit the `OrchestrationInstance` name from the methods

source/ProcessManager.Client/Processes/BRS_023_027/V1/INotifyAggregatedMeasureDataClientV1.cs
```diff
@@ -0,0 +1,54 @@
...
+        ScheduleOrchestrationInstanceDto<NotifyAggregatedMeasureDataInputV1> requestDto,
+        CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Get information for BRS-023 or BRS-027 calculation orchestration instance.
+    /// </summary>
+    public Task<OrchestrationInstanceTypedDto<NotifyAggregatedMeasureDataInputV1>> GetCalculationOrchestrationInstanceAsync(
```
ebbeknudsen, line 36:
```suggestion
    public Task<OrchestrationInstanceTypedDto<NotifyAggregatedMeasureDataInputV1>> GetCalculationAsync(
```

source/ProcessManager.Client/Processes/BRS_023_027/V1/INotifyAggregatedMeasureDataClientV1.cs
```diff
@@ -0,0 +1,54 @@
...
+        Guid id,
+        CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Get all BRS-023 or BRS-027 calculation orchestration instances filtered by given parameters.
+    /// </summary>
+    public Task<IReadOnlyCollection<OrchestrationInstanceTypedDto<NotifyAggregatedMeasureDataInputV1>>> SearchCalculationOrchestrationInstancesAsync(
```
ebbeknudsen, line 43:
```suggestion
    public Task<IReadOnlyCollection<OrchestrationInstanceTypedDto<NotifyAggregatedMeasureDataInputV1>>> SearchCalculationsAsync(
```

ebbeknudsen (dismissed):
LGTM 😉

source/ProcessManager.Client.Tests/Fixtures/ScenarioAppFixturesConfiguration.cs
dstenroejl: Its for sure not beautiful, and I would prefer to refactor it to a "settings" object that gets injected in the constructor, and then each fixture using it could call a Dispose on the full settings object (implemented to allow a call enve if it was already disposed). but I don't think there is an issue at the moment.

source/ProcessManager.Client.Tests/Integration/MonitorCalculationScenarioUsingApi.cs
dstenroejl: When comparing to Wholesale scenario tests we don't suffix them with tests. I'm not sure what to do here, but I believe we should have a team chat about naming tests anyways. For now I will rename so they are suffixed with Scenario.

ebbeknudsen (approved):
LGTM

PR #1337, opened 2024-11-06
feat: Update period validation for requests

MWO1024:
## Description

The updated rule for how far back one can request data is: 3.5 years back, with the entirety of the last month available. Assume today is November 15, 2024, then the following (aggregation) requests are allowed:

1. 2021, May 1--7
2. 2021, May 8--15
3. 2021, May 10-20
4. 2021, May 17-24
5. 2021, May 1--31 

## Checklist
- [ ] Reference to the task


source/Edi.UnitTests/Validators/AggregatedTimeSeriesRequest/PeriodValidatorTests.cs
```diff
@@ -230,4 +242,90 @@ public async Task Validate_WhenPeriodOverlapWinterDaylightSavingTime_ReturnsNoVa
...
+
+    [Fact]
+    public async Task Validate_WhenPeriodStartIsLessThan3YearsAnd6MonthsOld_ReturnNoValidationError()
+    {
+        // Arrange
+        _now = new LocalDateTime(2024, 6, 1, 0, 0, 0)
+            .InZoneStrictly(_dateTimeZone!)
+            .ToInstant();
+
+        var start = new LocalDateTime(2022, 4, 1, 0, 0, 0)
```
RasmusGraabaek, line 274: Would it be beneficial to have start as
var start = _now.PlusYears(-2).PlusMonths(-2)

And so on?

RasmusGraabaek (approved):

source/Edi.UnitTests/Validators/AggregatedTimeSeriesRequest/PeriodValidatorTests.cs
MWO1024: I implemented the solution from the discussion. But I added comments instead of using variable names. It seemed neater in this context.

RasmusGraabaek (approved):

PR #1343, opened 2024-11-08
feat: Use current GAO as CO for RSM-019

MWO1024:
## Description

We have an issue where we sent wholesale services to the old charge owner instead of the current grid owner. 

The flows are illustrated here: https://miro.com/app/board/uXjVMCOTRs0=/?moveToWidget=3458764606456331549&cot=14
where the green post-it shows the agreed changes to fix this issue. 

When this has been merged the drawing will be updated and the green post-its will be removed

## Checklist
- [ ] Reference to the task


source/MasterData.Application/MasterDataClient.cs
```diff
@@ -109,6 +110,21 @@ public async Task<GridAreaOwnerDto> GetGridOwnerForGridAreaCodeAsync(string grid
         return new GridAreaOwnerDto(gridAreaOwner.GridAreaCode, gridAreaOwner.ValidFrom, gridAreaOwner.GridAreaOwnerActorNumber, gridAreaOwner.SequenceNumber);
     }
 
+    public async IAsyncEnumerable<GridAreaOwnerDto> GetAllGridAreaOwnersAsync(
```
lasrinnil, line 113: I dont see the benefit os using a `IAsyncEnumerable` in this case, the usage of the collection is applied only after the whole collection has been materialized.

source/B2BApi/Functions/EnqueueMessages/Activities/GetGridAreaOwnersActivity.cs
```diff
@@ -0,0 +1,54 @@
...
+        TaskOptions? options)
+    {
+        return context.CallActivityAsync<ImmutableDictionary<string, ActorNumber>>(
+            nameof(GetGridAreaOwnersActivity),
+            null,
+            options: options);
+    }
+
+    [Function(nameof(GetGridAreaOwnersActivity))]
+    public async Task<ImmutableDictionary<string, ActorNumber>> Run([ActivityTrigger] object? input)
```
lasrinnil, line 43: Why not return collection of `GridAreaOwnerDto`s ?

source/OutgoingMessages.Infrastructure/Databricks/WholesaleResults/Queries/WholesaleAmountPerChargeQuery.cs
```diff
@@ -85,13 +87,25 @@ protected override async Task<WholesaleAmountPerChargeMessageDto> CreateWholesal
 
         if (isTax)
         {
-            var gridAreaOwner = await _masterDataClient
-                .GetGridOwnerForGridAreaCodeAsync(gridAreaCode, CancellationToken.None)
-                .ConfigureAwait(false);
+            var gridAreaOwner = _gridAreaOwners[gridAreaCode];
```
lasrinnil, line 90: The way I am reading this, the `chargeOwnerReceiverId` is overwritten with the `gridAreaOwner` if either `IsTax` is `true` or `false`, we just log if the initial `chargeOwnerId` is not the `SystemOperatorActorNumber`. Is this the correct behavior?

source/OutgoingMessages.Infrastructure/Databricks/WholesaleResults/Queries/WholesaleMonthlyAmountPerChargeQuery.cs
```diff
@@ -76,19 +79,32 @@ protected override async Task<WholesaleMonthlyAmountPerChargeMessageDto> CreateW
 
         if (isTax)
         {
-            var gridAreaOwner = await _masterDataClient
-                .GetGridOwnerForGridAreaCodeAsync(gridAreaCode, CancellationToken.None)
-                .ConfigureAwait(false);
+            var gridAreaOwner = _gridAreaOwners[gridAreaCode];
```
lasrinnil, line 82: please see my comment on `WholesaleAmountPerChargeQuery`

source/B2BApi/Functions/EnqueueMessages/Activities/GetGridAreaOwnersActivity.cs
MWO1024: Then we would have to search the list each time (giving us a O(n) lookup), instead of the approx O(1) lookup of a dictionary (they are implemented as hash tables; see CLRS for more info). Due to the fact we have to lookup the grid area owner for every single row in databricks, it would be inadvisable to not use a proper data structure (not to mention it is just a bad design in general).

source/OutgoingMessages.Infrastructure/Databricks/WholesaleResults/Queries/WholesaleAmountPerChargeQuery.cs
MWO1024: This is correct. The receiver is either EZ or the GAO, see 
billede

MWO1024: We log the cases where we use another receiver than the charge owner, so we have a chance to debug/preserve a trace.

source/OutgoingMessages.Infrastructure/Databricks/WholesaleResults/Queries/WholesaleTotalAmountQuery.cs
```diff
@@ -103,17 +106,24 @@ protected override Task<WholesaleTotalAmountMessageDto> CreateWholesaleResultAsy
...
+        var gridAreaCode = databricksSqlRow.ToNonEmptyString(WholesaleResultColumnNames.GridAreaCode);
         var energySupplierNumber =
             ActorNumber.Create(databricksSqlRow.ToNonEmptyString(WholesaleResultColumnNames.EnergySupplierId));
+
         if (chargeOwnerNumber is not null)
         {
+            if (chargeOwnerNumber != DataHubDetails.SystemOperatorActorNumber)
+            {
+                var gridAreaOwner = gridAreaOwnerDictionary[gridAreaCode];
+                return (gridAreaOwner, GetChargeOwnerRole(gridAreaOwner));
```
lasrinnil, line 122:
```suggestion
                return (gridAreaOwner, ActorRole.GridAccessProvider);
```

Would it be more explicit that in the if case that the actor is not the `SystemOperatorActorNumber` it must be a GridAccessProvider?

I would suggest we remove `GetChargeOwnerRole()` to make the code easier to read.

lasrinnil (dismissed):

source/MasterData.Application/MasterDataClient.cs
```diff
@@ -110,19 +111,17 @@ public async Task<GridAreaOwnerDto> GetGridOwnerForGridAreaCodeAsync(string grid
...
-                           .ConfigureAwait(false))
+        var builder = ImmutableDictionary.CreateBuilder<string, ActorNumber>();
+        await foreach (var gridAreaOwner in _gridAreaRepository.GetAllGridAreaOwnersAsync(cancellationToken))
         {
-            yield return new GridAreaOwnerDto(
-                gridAreaOwner.GridAreaCode,
-                gridAreaOwner.ValidFrom,
-                gridAreaOwner.GridAreaOwnerActorNumber,
-                gridAreaOwner.SequenceNumber);
+            // TODO (MWO): This should in principle always add, and always successfully add, but we should consider what to do if it fails.
```
lasrinnil, line 120: Planning to address this comment in this PR or an upcoming one? Perhaps we should raise an exception if we are unable to create a valid set of grid area owners.

source/B2BApi/Functions/EnqueueMessages/Activities/GetGridAreaOwnersActivity.cs
```diff
@@ -40,15 +40,8 @@ public static Task<ImmutableDictionary<string, ActorNumber>> StartActivityAsync(
...
     {
-        var builder = ImmutableDictionary.CreateBuilder<string, ActorNumber>();
-        await foreach (var (gridAreaCode, _, actorNumber, _) in _masterDataClient.GetAllGridAreaOwnersAsync(
-                           CancellationToken.None))
-        {
-            builder.TryAdd(gridAreaCode, actorNumber);
-        }
-
-        return builder.ToImmutable();
+        return _masterDataClient.GetAllGridAreaOwnersAsync(CancellationToken.None);
```
lasrinnil, line 45: I think it would serve the `MasteDataClient` best to expose the `GridAreaOwners` model, but the activity would create a `ImmutableDictionary<string, ActorNumber>` to best serve its usecase.

source/OutgoingMessages.Infrastructure/Databricks/WholesaleResults/Queries/WholesaleAmountPerChargeQuery.cs
```diff
@@ -85,26 +85,15 @@ protected override Task<WholesaleAmountPerChargeMessageDto> CreateWholesaleResul
         var chargeOwnerReceiverId = originalChargeOwnerReceiverId;
         var isTax = databricksSqlRow.ToBool(WholesaleResultColumnNames.IsTax);
 
-        if (isTax)
+        if (isTax || chargeOwnerId != DataHubDetails.SystemOperatorActorNumber)
```
lasrinnil, line 88: 👍

lasrinnil: nice

lasrinnil (approved):

PR #1377, opened 2024-11-19
feat: new incoming rsm-012 json message parser

MadsDue:
## Description
This PR contains a new JSON parser for metered data for a measurement point,

## Checklist
- [ ] Reference to the task


source/IncomingMessages.Infrastructure/MessageParsers/MeteredDateForMeasurementPointParsers/Ebix/MeteredDataForMeasurementPointEbixMessageParser.cs
```diff
@@ -28,6 +28,8 @@ public async Task<IncomingMarketMessageParserResult> ParseAsync(
         IIncomingMarketMessageStream incomingMarketMessageStream,
         CancellationToken cancellationToken)
     {
-        return await _messageParser.ParseAsync(incomingMarketMessageStream, cancellationToken).ConfigureAwait(false);
+        var messageParser = _messageParsers
+            .Single(x => x.DocumentFormat == DocumentFormat.Ebix && x.DocumentType == DocumentType);
+        return await messageParser.ParseAsync(incomingMarketMessageStream, cancellationToken).ConfigureAwait(false);
```
MWO1024, line 33: I wonder if we could utilise keyed services and/or a dictionary here as well. The keyed services to only get the ebix ones in the constructor, and then use the dictionary to improve document type lookup. Not sure if it is worth it.

MWO1024 (approved):

source/IncomingMessages.Infrastructure/MessageParsers/MeteredDateForMeasurementPointParsers/Ebix/MeteredDataForMeasurementPointEbixMessageParser.cs
MadsDue: This is the old pattern and will be remove shortly :)

PR #1378, opened 2024-11-19
refac: Store identity info for Process Manger

dstenroejl:
## Description

- [x] Add UserIdentity, ActorIdentity and OperatingIdentity
- [x] Extend orchestration instance with information about who created and canceled it

## Checklist
- [ ] Reference to the task


source/ProcessManager.DatabaseMigration/Scripts/202411192300 Add operating identity columns.sql
```diff
@@ -0,0 +1,8 @@
+﻿ALTER TABLE [pm].[OrchestrationInstance]
```
dstenroejl, line 1: - Should we migrate any data, or can we skip that since we will create a new database in the new repository?
- Also, should we require `Lifecycle_CreatedBy_IdentityType` by setting it `NOT NULL`?

source/ProcessManager.Client.Tests/Fixtures/ScenarioOrchestrationsAppFixture.cs
```diff
@@ -31,7 +31,8 @@ public ScenarioOrchestrationsAppFixture()
         : base(
             ScenarioAppFixturesConfiguration.Instance.DatabaseManager,
             ScenarioAppFixturesConfiguration.Instance.TaskHubName,
-            ScenarioAppFixturesConfiguration.Instance.OrchestrationsAppPort)
+            ScenarioAppFixturesConfiguration.Instance.OrchestrationsAppPort,
+            disposeDatabase: true) // Quickfix: Only dispose database in one of the scenario fixtures
```
dstenroejl, line 35: CI started to fail a lot with an error that could be related to running DatabaseManage.Dispose() simultaneously - so I added a quick-fix. I did not want to change too much because Ebbe was talking about refactoring the fixtures in a separate PR.

source/ProcessManager.Client.Tests/Integration/MonitorCalculationUsingApiScenario.cs
```diff
@@ -69,14 +69,16 @@ public async Task CalculationBrs023_WhenScheduledUsingClient_CanMonitorLifecycle
...
+        scheduleRequestDto.OperatingIdentity.UserId = Guid.NewGuid();
+        scheduleRequestDto.OperatingIdentity.ActorId = Guid.NewGuid();
         scheduleRequestDto.RunAt = "2024-11-01T06:19:10.0209567+01:00";
         scheduleRequestDto.InputParameter = new ExpandoObject();
         scheduleRequestDto.InputParameter.CalculationType = 0;
         scheduleRequestDto.InputParameter.GridAreaCodes = new[] { "543" };
         scheduleRequestDto.InputParameter.PeriodStartDate = "2024-10-29T15:19:10.0151351+01:00";
         scheduleRequestDto.InputParameter.PeriodEndDate = "2024-10-29T16:19:10.0193962+01:00";
         scheduleRequestDto.InputParameter.IsInternalCalculation = true;
-        scheduleRequestDto.InputParameter.UserId = Guid.NewGuid();
```
ebbeknudsen, line 79: Shouldn't this be able to use the request dto the API accepts?

source/ProcessManager.Client/Processes/BRS_026_028/V1/RequestCalculatedDataClientV1.cs
```diff
@@ -30,47 +30,47 @@ public class RequestCalculatedDataClientV1(
...
-    where TInput : class
+        MessageCommand<TInputParameterDto> inputParameter)
+    where TInputParameterDto : IInputParameterDto
     {
         var message = new StartOrchestrationDto
         {
             OrchestrationName = orchestrationName,
             OrchestrationVersion = orchestrationVersion,
-            JsonInput = JsonSerializer.Serialize(input.Input),
+            JsonInput = JsonSerializer.Serialize(inputParameter.InputParameter),
```
ebbeknudsen, line 67:
```suggestion
            JsonInput = JsonSerializer.Serialize(command.InputParameter),
```

source/ProcessManager.DatabaseMigration/Scripts/202411192300 Add operating identity columns.sql
ebbeknudsen: I would prefer that the datamodel is modelled as what we think is correct (so non null `IdentityType` and maybe `ActorId`). We've done that by doing something like this in our SQL scripts:

```sql

ALTER TABLE [pm].[OrchestrationInstance]
    ADD [IdentityType] NVARCHAR(255) NULL
GO

UPDATE [pm].[OrchestrationInstance]
   SET [IdentityType] = 'NULL-VALUE-SET-BY-SCRIPT' -- <-- Use a value that makes it easy to determine the value is set by this script in the future
   WHERE [IdentityType] IS NULL
GO

ALTER TABLE [pm].[OrchestrationInstance]
    ALTER COLUMN [IdentityType] NVARCHAR(255) NOT NULL
GO
```

ebbeknudsen (approved):

ebbeknudsen (approved):

ebbeknudsen (approved):

PR #1380, opened 2024-11-21
refac: 3. refactor outgoing messages to clean architecture

AndersBallingPetersen:
## Description
This module adheres to the illustration below.

image
## Checklist
- [ ] Reference to the task


source/OutgoingMessages.Infrastructure/OutgoingMessages.Infrastructure.csproj
```diff
@@ -7,9 +7,6 @@
   
   <ItemGroup>
     <PackageReference Include="Energinet.DataHub.Core.Databricks.SqlStatementExecution" Version="11.2.4" />
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
```
MadsDue, line 10: Is does not need anymore?

AndersBallingPetersen: This package is unused. All async features comes out of the box after .NET 5. This should be included if we support .NET 2 and prior.

PR #1381, opened 2024-11-21
fix: bug where actorRole for archived message is null

MadsDue:
## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/11970919288
- [ ] Reference to the task


ManBearTM (approved):
LGTM!

MWO1024 (approved):

PR #1383, opened 2024-11-21
refac: Refactor PM shared test fixtures

ebbeknudsen:
## Description

Refactor Process Manager tests to share and use managers for running the different apps, instead of sharing and using multiple fixture.

**Before**
- `MonitorCalculationUsingClientsScenario`
    - `ScenarioProcessManagerAppFixture` (shared)
    - `OrchestrationsAppFixture` (shared)

**After**
- `MonitorCalculationUsingClientsScenario`
    - `ProcessManagerClientFixture`
        - `ProcessManagerAppManager` (shared)
        - `OrchestrationsAppManager` (shared)

## Checklist
- [ ] Reference to the task

# COPILOT SUMMARY 🤣 🔥 

This pull request involves significant changes to the test fixtures and integration tests in the `ProcessManager.Client.Tests` project. The primary focus is on consolidating and simplifying the test setup by introducing a new `ProcessManagerClientFixture` and removing redundant fixtures.

### Test Fixture Consolidation:
* [`ProcessManagerClientCollection.cs`](diffhunk://#diff-df01fad5de2e5370d0c9c26f286b6095be662406591a9590a1512fd4f526e44dL27-R27): Updated to use `ProcessManagerClientFixture` instead of `ScenarioProcessManagerAppFixture` and `ScenarioOrchestrationsAppFixture`.
* [`ProcessManagerClientFixture.cs`](diffhunk://#diff-bdf0a6f17431fb8e0476b5db11cbf2877ef75f1094ce1f3970538c141a0719c8R1-R104): Introduced a new fixture to manage common test setup tasks, including initializing and disposing resources.

### Removal of Redundant Fixtures:
* `ScenarioAppFixturesConfiguration.cs`, `ScenarioOrchestrationsAppFixture.cs`, and `ScenarioProcessManagerAppFixture.cs`: Removed these redundant fixtures as their responsibilities are now handled by `ProcessManagerClientFixture`. [[1]](diffhunk://#diff-bad3fb3bf54cd967159d27d2b140af7921a10b4700755a98cb476aaf532f2ef2L1-L51) [[2]](diffhunk://#diff-ead979995b7fb1402e9d48fc4caeb6a4a8e2ba28b5f35a7fdfc889c10920244cL1-L38) [[3]](diffhunk://#diff-adc8f85deb95c7ec9f92d02d2779e0d35b4b0e56ec35614a4db333546ff6b756L1-L38)

### Integration Test Updates:
* [`RequestCalculatedEnergyTimeSeriesTests.cs`](diffhunk://#diff-8c749ba22de20c63863c20c32888f7a96e7d4207c2d4c1c67cd2bf92b23fbbc3R15-R25): Updated to use `ProcessManagerClientFixture` and refactored the test to use the new fixture setup. [[1]](diffhunk://#diff-8c749ba22de20c63863c20c32888f7a96e7d4207c2d4c1c67cd2bf92b23fbbc3R15-R25) [[2]](diffhunk://#diff-8c749ba22de20c63863c20c32888f7a96e7d4207c2d4c1c67cd2bf92b23fbbc3L27-R86)
* `MonitorCalculationUsingApiScenario.cs` and `MonitorCalculationUsingClientsScenario.cs`: Updated to use `ProcessManagerClientFixture` and refactored methods to align with the new fixture. [[1]](diffhunk://#diff-d97d7fcc821b94030468ca38e768e8f010ec7e36aa8d388748bd8582d78c6b63L36-R55) [[2]](diffhunk://#diff-d97d7fcc821b94030468ca38e768e8f010ec7e36aa8d388748bd8582d78c6b63L92-R85) [[3]](diffhunk://#diff-d97d7fcc821b94030468ca38e768e8f010ec7e36aa8d388748bd8582d78c6b63L101-R94) [[4]](diffhunk://#diff-d97d7fcc821b94030468ca38e768e8f010ec7e36aa8d388748bd8582d78c6b63L112-R105) [[5]](diffhunk://#diff-fe1542a4d55aca306cd3765dea317518f88dc8872a250ce3185c26babdac1ad9L43-R75) [[6]](diffhunk://#diff-fe1542a4d55aca306cd3765dea317518f88dc8872a250ce3185c26babdac1ad9L109-R102)

### Project File Updates:
* [`ProcessManager.Client.Tests.csproj`](diffhunk://#diff-fce0ca8af11656bf8be02aeb5d8e84ec5d9c53f0af075165315375f092d6baedL12-R17): Updated to include new fixture files and remove references to old fixture base classes.


docs/ProcessManager.Client/ReleaseNotes/ReleaseNotes.md
```diff
@@ -1,5 +1,9 @@
 # ProcessManager.Client Release Notes
 
+## Version 0.11.1
+
+- No functional changes.
+
```
ebbeknudsen, line 6: Should `ProcessManager.Client.Tests` changes really result in a new NuGet package release?

source/ProcessManager.Client/ProcessManager.Client.csproj
```diff
@@ -7,7 +7,7 @@
 
   <PropertyGroup>
     <PackageId>Energinet.DataHub.ProcessManager.Client</PackageId>
-    <PackageVersion>0.11.0$(VersionSuffix)</PackageVersion>
+    <PackageVersion>0.11.1$(VersionSuffix)</PackageVersion>
     <Title>DH3 Process Manager Client library</Title>
```
ebbeknudsen, line 11: Should `ProcessManager.Client.Tests` changes really require a new NuGet package release?

source/ProcessManager.Client.Tests/Fixtures/ProcessManagerClientFixture.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        AzuriteManager = new AzuriteManager(useOAuth: true);
+
+        OrchestrationsAppManager = new OrchestrationsAppManager(
+            DatabaseManager,
+            IntegrationTestConfiguration,
+            AzuriteManager,
+            taskHubName,
+            8101,
+            false);
```
dstenroejl, line 45: When we use numbers and booleans in parameters consider to show the parameter name, as there is no "context" so adding it makes it easier to review.
```suggestion
            port: 8101,
            manageAzurite: false);
```

source/ProcessManager.Client.Tests/Fixtures/ProcessManagerClientFixture.cs
```diff
@@ -0,0 +1,104 @@
...
+        ProcessManagerAppManager = new ProcessManagerAppManager(
+            DatabaseManager,
+            IntegrationTestConfiguration,
+            AzuriteManager,
+            taskHubName,
+            8102,
+            false);
```
dstenroejl, line 52:
```suggestion
            port: 8102,
            manageAzurite: false);
```

source/ProcessManager.Tests/Fixtures/ProcessManagerAppManager.cs
```diff
@@ -75,13 +88,16 @@ public ProcessManagerAppFixtureBase(
...
+            CleanupAzuriteStorage();
 
-        // Storage emulator
-        AzuriteManager.StartAzurite();
+            // Storage emulator
+            AzuriteManager.StartAzurite();
+        }
 
         // Database
         await DatabaseManager.CreateDatabaseAsync();
```
dstenroejl, line 103: Maybe we should also have a parameter for "manage Database", since I had to fix it in CI in EDI, so we will most likely just have to do it at some point later anyway?

Same comment for `OrchestrationsAppManager`

source/ProcessManager.Orchestrations.Tests/Fixtures/OrchestrationsAppManager.cs
```diff
@@ -14,45 +14,70 @@
...
 
         HostConfigurationBuilder = new FunctionAppHostConfigurationBuilder();
+
+        ServiceBusResourceProvider = new ServiceBusResourceProvider(
+            TestLogger,
+            IntegrationTestConfiguration.ServiceBusFullyQualifiedNamespace,
+            IntegrationTestConfiguration.Credential);
     }
 
+    public ServiceBusResourceProvider ServiceBusResourceProvider { get; }
```
dstenroejl, line 79: Can be private. The intention is that we use the provider to create Topics / Subscription and they can be public, but the provider should be kept private and not accessed from the outside.

Same comment for `ProcessManagerClientFixture`

source/ProcessManager.Client.Tests/Fixtures/ProcessManagerClientFixture.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        ServiceBusResourceProvider = new ServiceBusResourceProvider(
+            OrchestrationsAppManager.TestLogger,
+            IntegrationTestConfiguration.ServiceBusFullyQualifiedNamespace,
+            IntegrationTestConfiguration.Credential);
+    }
+
+    public AzuriteManager AzuriteManager { get; }
```
dstenroejl, line 60: Can be private

source/ProcessManager.Client.Tests/Integration/BRS_026_028/V1/RequestCalculatedEnergyTimeSeriesTests.cs
```diff
@@ -24,51 +34,56 @@ namespace Energinet.DataHub.ProcessManager.Client.Tests.Integration.BRS_026_028.
...
-        // var requestCalculatedDataClient = new RequestCalculatedDataClientV1();
-        // var input = new RequestCalculatedDataInputV1<RequestCalculatedEnergyTimeSeriesInputV1>(
-        //     Guid.NewGuid().ToString(),
-        //     new RequestCalculatedEnergyTimeSeriesInputV1("B1337"));
-        //
-        // // Act
-        // await requestCalculatedDataClient.RequestCalculatedEnergyTimeSeriesAsync(input, CancellationToken.None);
+        await using var serviceBusClient = new ServiceBusClient(
+            Fixture.IntegrationTestConfiguration.ServiceBusFullyQualifiedNamespace,
+            Fixture.IntegrationTestConfiguration.Credential);
```
dstenroejl, line 68: No! Create the TopicResource and make it public on the fixture (as a property), then it has a `SenderClient` property that you should use. The client is lazy created, and automatically disposed (if it was ever created). :)

dstenroejl (commented):
Commented

source/ProcessManager.Client.Tests/Integration/BRS_026_028/V1/RequestCalculatedEnergyTimeSeriesTests.cs
ebbeknudsen: Thanks! 👍

source/ProcessManager.Tests/Fixtures/ProcessManagerAppManager.cs
ebbeknudsen: I'm not sure what the exact problem was in EDI? I've looked at the code for creating/disposing the database, and as far as i can see it should all work if the database is already created/disposed. Maybe when using 2 fixtures initialize/dispose happened in parallel (which is no longer the case), which was why an error happened?

But it could still make sense for clarity/performance reasons to only do it once.

source/ProcessManager.Tests/Fixtures/ProcessManagerAppManager.cs
```diff
@@ -25,31 +25,44 @@
...
-        _disposeDatabase = disposeDatabase;
 
         TestLogger = new TestDiagnosticsLogger();
-        IntegrationTestConfiguration = new IntegrationTestConfiguration();
+        IntegrationTestConfiguration = integrationTestConfiguration;
 
-        AzuriteManager = new AzuriteManager(useOAuth: true);
+        AzuriteManager = azuriteManager;
 
         HostConfigurationBuilder = new FunctionAppHostConfigurationBuilder();
```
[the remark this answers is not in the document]
ebbeknudsen: The Azurite manager doesn't contain any state (which you only can see if you inspect the `AzuriteManager` code), so it could just be used as `new()` everywhere it is used. However that is not the pattern we usually use, and it also requires the developer to look at the `AzuriteManager` implementation (which is in another repository) to understand why we don't have to keep the reference. I don't think that is a good tradeoff just to not save the AzuriteManager in a variable, and would also break if the `AzuriteManager` is ever updated to contain any state. Maybe i misundstood your intention?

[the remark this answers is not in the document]
dstenroejl: I mean the `HostConfigurationBuilder` property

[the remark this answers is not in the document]
ebbeknudsen: That makes a lot more sense than the `AzuriteManager` 😄 . I'll update it.

dstenroejl (approved):
LGTM

PR #1384, opened 2024-11-22
fix: outdated eic function contract

MadsDue:
## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/11969808683
- [ ] Reference to the task


dstenroejl (approved):
LGTM

PR #1409, opened 2024-12-03
refac: move incoming message validation requirement into domain models

MadsDue:
## Description

I refactored the IIncomingMessage interface to include properties for specifying allowed values for business validation in the sync flow:

- IReadOnlyCollection<string> AllowedMessageTypes { get; }
- IReadOnlyCollection<BusinessReason> AllowedBusinessReasons { get; }
- IReadOnlyCollection<ActorRole> AllowedSenderRoles { get; }

With this refactoring, any new implementation of IIncomingMessage must define the allowed values for these properties. The validation framework will then automatically validate the document against these values. As a result, it is no longer necessary to extend the validation framework when introducing a new IIncomingMessage implementation.

## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/12176796485
- [ ] Reference to the task

### DIFF
--- a/source/IncomingMessages.Domain/Abstractions/IIncomingMessage.cs
+++ b/source/IncomingMessages.Domain/Abstractions/IIncomingMessage.cs
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Energinet.DataHub.EDI.BuildingBlocks.Domain.DataHub;
+using Energinet.DataHub.EDI.BuildingBlocks.Domain.Models;
+
 namespace Energinet.DataHub.EDI.IncomingMessages.Domain.Abstractions;
 
 /// <summary>
@@ -68,4 +71,10 @@ public interface IIncomingMessage
     /// Series of the incoming message
     /// </summary>
     IReadOnlyCollection<IIncomingMessageSeries> Series { get; }
+
+    public IReadOnlyCollection<string> AllowedMessageTypes { get; }
+
+    public IReadOnlyCollection<BusinessReason> AllowedBusinessReasons { get; }
+
+    public IReadOnlyCollection<ActorRole> AllowedSenderRoles { get; }
 }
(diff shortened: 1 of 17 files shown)


source/IncomingMessages.Domain/Validation/SenderAuthorizer.cs
```diff
@@ -68,59 +62,26 @@ private void EnsureSenderRoleCode(IIncomingMessage message, bool allSeriesAreDel
...
-                        break;
-                    default:
-                        _validationErrors.Add(new SenderRoleTypeIsNotAuthorized());
-                        break;
-                }
-
-                break;
-            default:
-                throw new ArgumentOutOfRangeException(nameof(message));
+            _validationErrors.Add(new SenderRoleTypeIsNotAuthorized());
```
RasmusGraabaek, line 73: source/IncomingMessages.Domain/Validation/ProcessTypeValidator.cs and the if statement above is like:
If (something ) => no error.

Where this statement is if (something) => error

Would it be beneficial to align these?

source/IncomingMessages.Domain/Abstractions/IIncomingMessage.cs
```diff
@@ -68,4 +71,10 @@ public interface IIncomingMessage
     /// Series of the incoming message
     /// </summary>
     IReadOnlyCollection<IIncomingMessageSeries> Series { get; }
+
+    public IReadOnlyCollection<string> AllowedMessageTypes { get; }
```
RasmusGraabaek, line 75: I do not like the fact, that "AllowedMessageTypes" is a string array
But I'm pretty sure, that we're unable to give, eg. "E66" a good name :(

MadsDue: We can make a type of it.

RasmusGraabaek (approved):

PR #1413, opened 2024-12-06
feat: Add document validation to RSM-12 behaviour tests

MWO1024:
## References

https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/407

## Checklist
- [x] Reference to the task


source/IntegrationTests/Behaviours/IncomingRequests/GivenMeteredDataForMeasurementPointTests.cs
```diff
@@ -77,10 +83,50 @@ await GivenReceivedMeteredDataForMeasurementPoint(
         // Assert
         foreach (var peekResultDto in peekResults)
```
MadsDue, line 84: Should we assert that the peekResults contains minimum 2 results?

source/OutgoingMessages.IntegrationTests/DocumentAsserters/NotifyValidatedMeasureDataDocumentAssertionInput.cs
```diff
@@ -12,8 +12,43 @@
...
+        RequiredSeriesFields = requiredSeriesFields;
+    }
+
+    public RequiredDocumentFields RequiredDocumentFields { get; }
+
+    public RequiredSeriesFields? RequiredSeriesFields { get; }
 }
+
+public sealed record RequiredSeriesFields(
+    TransactionId TransactionId,
```
MadsDue, line 46: I think "originalTransactionIDReference_Series.mRID" is a required RSM012 field for outgoing RSM012.

source/Tests/Infrastructure/OutgoingMessages/RSM012/IAssertMeteredDateForMeasurementPointDocumentDocument.cs
```diff
@@ -12,14 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System.Collections.ObjectModel;
```
MadsDue, line 15: Í dont think this is needed

source/OutgoingMessages.Domain/DocumentWriters/RSM012/MeteredDateForMeasurementPointCimJsonDocumentWriter.cs
```diff
@@ -100,7 +100,7 @@ private Document ParseFrom(OutgoingMessageHeader header, IReadOnlyCollection<str
                 header.SenderId,
                 header.SenderRole,
                 TypeCode,
-                meteredDataForMeasurementPoints));
+                meteredDataForMeasurementPoints.Count == 0 ? null : meteredDataForMeasurementPoints));
```
MadsDue, line 103: There is always point, right?

MWO1024: In this case, the type is defined as
```internal class MeteredDataForMeasurementPoint(
    string transactionId,
    string marketEvaluationPointNumber,
    string marketEvaluationPointType,
    string? originalTransactionIdReferenceId,
    string? product,
    string quantityMeasureUnit,
    string? registrationDateTime,
    Period period)```
which is a series. These we don't need (according to the schemes). The `Period` type then contains what we would usually think of as points:
```internal class Period(string resolution, TimeInterval timeInterval, IReadOnlyCollection<Point> points)
{
    [JsonPropertyName("resolution")]
    public string Resolution { get; init; } = resolution;

    [JsonPropertyName("timeInterval")]
    public TimeInterval TimeInterval { get; init; } = timeInterval;

    [JsonPropertyName("Point")]
    public IReadOnlyCollection<Point> Points { get; init; } = points;
}```

Comment (MWO1024) [reply source/OutgoingMessages.Domain/DocumentWriters/RSM012/MeteredDateForMeasurementPointCimJsonDocumentWriter.cs:103]: I should probably rename the first type to `Series` instead of `Point` to make it less confusing

File: source/OutgoingMessages.Domain/DocumentWriters/RSM012/MeteredDateForMeasurementPointCimJsonDocumentWriter.cs [hunk 5]
```diff
@@ -150,17 +150,19 @@ internal class MeteredDateForMeasurementPoint(
     public ValueObject<string> Type { get; init; } = ValueObject<string>.Create(typeCode);
 
     [JsonPropertyName("Series")]
-    public IReadOnlyCollection<MeteredDataForMeasurementPoint> MeteredDataForMeasurementPoints { get; init; } = meteredDataForMeasurementPoints;
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
```
MadsDue, line 153: Same here, is there not always a Series?

source/Tests/Infrastructure/OutgoingMessages/RSM012/IAssertMeteredDateForMeasurementPointDocumentDocument.cs
```diff
@@ -13,13 +13,14 @@
...
-using Energinet.DataHub.EDI.OutgoingMessages.Domain.DocumentWriters.RSM012;
 
 namespace Energinet.DataHub.EDI.Tests.Infrastructure.OutgoingMessages.RSM012;
 
 public interface IAssertMeteredDateForMeasurementPointDocumentDocument
 {
-    IAssertMeteredDateForMeasurementPointDocumentDocument HasMessageId(string expectedMessageId);
+    IAssertMeteredDateForMeasurementPointDocumentDocument MessageIdExists();
+
+    IAssertMeteredDateForMeasurementPointDocumentDocument HasNoSeriesElements();
```
MadsDue, line 23: This should never happen, right?

source/OutgoingMessages.Domain/DocumentWriters/RSM012/MeteredDateForMeasurementPointCimJsonDocumentWriter.cs
MWO1024: I have reverted this `== 0` check, as we don't expect any empty RSM-12s

source/OutgoingMessages.Domain/DocumentWriters/RSM012/MeteredDateForMeasurementPointCimJsonDocumentWriter.cs
MWO1024: I have reverted this omision of `Series` despite it being optional. We do not expect to ever write a document in which it is not present.

source/Tests/Infrastructure/OutgoingMessages/RSM012/IAssertMeteredDateForMeasurementPointDocumentDocument.cs
MWO1024: I have reverted the addition of `HasNoSeriesElements`. I'm tempted to keep it in some form, to validate that we can in fact write en "empty" RSM-12 if the need arises.

MadsDue (approved):
LGTM

source/OutgoingMessages.IntegrationTests/DocumentAsserters/NotifyValidatedMeasureDataDocumentAssertionInput.cs
MWO1024: When we get a bit wiser, we should revisit this. Ideally I think if/when we split the model depending on source: received from a grid area operator, result of Xyz calculation, and whatever else we have.
For instance, I'm not sure what the original transaction id would be for the result of some calculation or if there even is one.

PR #1426, opened 2024-12-19
fix: Read Point from Period and not Series elements

MWO1024:
## Description

- We accidentally read the points from the `Series` element, instead of the `Period` for JSON.
- We assumed that a `Point` always had all its values for JSON. This is incorrect, as two of them are nullable.
- We created a lot of points for our behaviour tests. This is now reduced to four: one for each structural case.
- We did not handle decimals correctly wrt assertion and temporary serialisation.

### Observation
- `GivenIncomingMessagesWithDelegationTests` relies on eBix for RSM-12. It should work for XML/JSON too, but it does not. We should fix this in another PR.

## References

https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/377

## Checklist
- [x] Reference to the task

### DIFF
--- a/source/IncomingMessages.IntegrationTests/Builders/MeteredDataForMeasurementPointBuilder.cs
+++ b/source/IncomingMessages.IntegrationTests/Builders/MeteredDataForMeasurementPointBuilder.cs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Text;
 using System.Xml;
@@ -73,7 +72,8 @@ public static class MeteredDataForMeasurementPointBuilder
                 businessType,
                 messageId,
                 senderRole,
-                receiverNumber);
+                receiverNumber,
+                schema ?? "NotifyValidatedMeasureData_MarketDocument");
         }
         else
         {
@@ -200,10 +200,11 @@ public static class MeteredDataForMeasurementPointBuilder
         string businessType,
         string messageId,
         string senderRole,
-        string receiverNumber) =>
+        string receiverNumber,
+        string schema) =>
         $$"""
           {
-            "NotifyValidatedMeasureData_MarketDocument": {
+            "{{schema}}": {
               "mRID": "{{messageId}}",
               "businessSector.type": {
                 "value": "{{businessType}}"
--- a/source/IncomingMessages.IntegrationTests/Builders/MeteredDataForMeasurementPointBuilder.cs
+++ b/source/IncomingMessages.IntegrationTests/Builders/MeteredDataForMeasurementPointBuilder.cs
@@ -62,6 +62,18 @@ public static class MeteredDataForMeasurementPointBuilder
                 receiverNumber,
                 schema ?? "urn:ediel.org:measure:notifyvalidatedmeasuredata:0:1");
         }
+        else if (format == DocumentFormat.Json)
+        {
+            content = GetJson(
+                senderActorNumber,
+                series,
+                messageType,
+                processType,
+                businessType,
+                messageId,
+                senderRole,
+                receiverNumber);
+        }
         else
         {
             throw new ArgumentOutOfRangeException(nameof(format), format, "Unsupported document format");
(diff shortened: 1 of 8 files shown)


source/Tests/CimMessageAdapter/Messages/MeteredDataForMeasurementPointMessageParserTests/MessageParserTests.cs
```diff
@@ -122,7 +122,7 @@ public async Task Successfully_parsed(DocumentFormat format, Stream message)
                 energyObservation.Position.Should().Be(position.ToString());
                 energyObservation.EnergyQuantity.Should().NotBeEmpty();
                 energyObservation.QuantityQuality.Should()
-                    .Match(quality => quality == "E01" || quality == "56" || quality == "A03");
+                    .Match(quality => quality == null || quality == "E01" || quality == "56" || quality == "A03" || quality == "A02" || quality == "A05");
```
lasrinnil, line 125: is `null` and `56` valid values for `Quality` in this case?

lasrinnil, line 125: ⚠️ This "OR" approach in test is very prone to bypass errors when input and output are inspected within a ranged.

lasrinnil (dismissed):

lasrinnil:
Please provide some more context on what the incorrect behavior was and the approach to solve it. This way the reviewer will be provided more insight quickly and would be able to give better feedback.

source/IncomingMessages.IntegrationTests/Builders/MeteredDataForMeasurementPointBuilder.cs
```diff
@@ -190,6 +202,94 @@ private static string GetXml(
...
+              "type": {
+          	     "value": "{{messageType}}"
+              },
+              "Series": [
+                {{string.Join(",\n", series.Select(s =>
+                    $$"""
+                      {
+                        "mRID": "{{s.TransactionId}}",
+                        "marketEvaluationPoint.mRID": {
+                        "codingScheme": "A10",
```
RasmusGraabaek, line 249: Small tab, por favor :D

source/IncomingMessages.IntegrationTests/Builders/MeteredDataForMeasurementPointBuilder.cs
```diff
@@ -190,6 +202,94 @@ private static string GetXml(
...
+                                $$"""
+                                  {
+                                    "position": {
+                                      "value": {{e.Position}}
+                                    },
+                                    "quality": {
+                                      "value": "A03"
+                                    },
+                                    "quantity": {{e.Quantity}}
+                                  }
```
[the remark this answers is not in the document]
MWO1024: Hmmm.... Yes. I think we do. I'll make a change for this for both XML and JSON. Just need to figure out how to sort-of-kinda randomise the result, without making it actually random; otherwise it is very hard to make an assert statement.

source/Tests/CimMessageAdapter/Messages/MeteredDataForMeasurementPointMessageParserTests/MessageParserTests.cs
RasmusGraabaek: As I undestand the json schema, the value 56 is not a valid quality value
image
The same is true for XML
image

But 56 is valid for ebix :(
image

RasmusGraabaek (approved):

source/Tests/CimMessageAdapter/Messages/MeteredDataForMeasurementPointMessageParserTests/MessageParserTests.cs
MWO1024: I have split the assertion in two: one for ebix and one for cim.

MWO1024: I agree. This probably isn't the best way to deal with this. It would probably be worth it, at some point, to sit down and figure out how best to test our parsers. I would think some sort of "best practise" exists for parsers; it is after all a very, very common tool.

source/IncomingMessages.IntegrationTests/Builders/MeteredDataForMeasurementPointBuilder.cs
```diff
@@ -21,6 +22,10 @@
 
 namespace Energinet.DataHub.EDI.IncomingMessages.IntegrationTests.Builders;
 
+[SuppressMessage(
```
lasrinnil, line 25: Why is this suppression required?

source/IncomingMessages.IntegrationTests/IncomingMessages/GivenIncomingMeteredDataForMeasurementMessageTests.cs
```diff
@@ -373,7 +373,7 @@ public async Task When_ProcessTypeIsNotAllowed_Then_ResultContainExceptedValidat
         result.Errors.Should().Contain(error => error is InvalidMessageStructure);
     }
 
-    [Theory]
+    [Theory(Skip = "We do not support eBix anymore")]
```
lasrinnil, line 376: Intended skip?

lasrinnil:
- Reverted 'RequestAndPeekFormatPairs' approch - while it is technically plausible for a actor to request in one format, and peek in another. It is not the inteded use, and not how the system is used by the actors. 


- Observation: if all new messages we are yet to support have the same amounts of integrationtests I worry for this test approach. 
	- Almost all of these tests have nothing to do with the specific messagetype under test.
	- Could we go with another approch to make sure that these tests would not have to be made for every type we support?

lasrinnil (approved):

PR #1428, opened 2024-12-23
refac: WhenArchivedMessageIsCreatedTests

lasrinnil:
## Description
This PR moves the `WhenArchivedMessageIsCreatedTests` from the "old" `IntegrationTest `project into the `ArchivedMessage.IntegrationTest`.
- Some tests have been rewritten to a simpler approach
- A single test have been removed.
   - This was dont because `ArchivedMessage`s FileStorageReference Id is now always made with a unique id when using the `ArchivedMessageDto`
image
 

## References
This change has no reference, its simply boyscout'ing 🫡 

## Checklist
- [x] Reference to the task


MadsDue (approved):
LGTM

PR #1453, opened 2025-01-22
feat: extend DataLakeFileStorageClient with a client to the new storage account

MadsDue:
## References
- infra: https://github.com/Energinet-DataHub/dh3-infrastructure/pull/2793

## Checklist
- [x] Subsystem test executed (dev_002/dev_003) https://github.com/Energinet-DataHub/dh3-environments/actions/runs/12949817090
- [x] Reference to the task

### DIFF
--- a/source/B2CWebApi.AppTests/Fixture/B2CWebApiFixture.cs
+++ b/source/B2CWebApi.AppTests/Fixture/B2CWebApiFixture.cs
@@ -126,6 +126,10 @@ public class B2CWebApiFixture : IAsyncLifetime
                 $"{BlobServiceClientConnectionOptions.SectionName}:{nameof(BlobServiceClientConnectionOptions.StorageAccountUrl)}",
                 AzuriteManager.BlobStorageServiceUri.AbsoluteUri
             },
+            {
+                $"{BlobServiceClientConnectionOptions.SectionName}:{nameof(BlobServiceClientConnectionOptions.StorageAccountUrlObsoleted)}",
+                AzuriteManager.BlobStorageServiceUri.AbsoluteUri
+            },
             { "UserAuthentication:MitIdExternalMetadataAddress", "YourMitIdExternalMetadataAddress" },
             { "UserAuthentication:ExternalMetadataAddress", OpenIdJwtManager.ExternalMetadataAddress },
             { "UserAuthentication:InternalMetadataAddress", OpenIdJwtManager.InternalMetadataAddress },
--- a/source/BuildingBlocks.Infrastructure/DataLakeFileStorageClient.cs
+++ b/source/BuildingBlocks.Infrastructure/DataLakeFileStorageClient.cs
@@ -28,20 +28,15 @@ namespace Energinet.DataHub.EDI.BuildingBlocks.Infrastructure;
 
 public class DataLakeFileStorageClient : IFileStorageClient
 {
-    private readonly IClock _clock;
     private readonly BlobServiceClient _blobServiceClientObsoleted;
     private readonly BlobServiceClient _blobServiceClient;
-    private readonly Instant _cutOffDate;
 
     public DataLakeFileStorageClient(
         IAzureClientFactory<BlobServiceClient> clientFactory,
-        IOptions<BlobServiceClientConnectionOptions> options,
-        IClock clock)
+        IOptions<BlobServiceClientConnectionOptions> options)
     {
-        _clock = clock;
         _blobServiceClientObsoleted = clientFactory.CreateClient(options.Value.ClientNameObsoleted);
         _blobServiceClient = clientFactory.CreateClient(options.Value.ClientName);
-        _cutOffDate = InstantPattern.General.Parse(options.Value.CutOffDate).Value;
     }
 
     public async Task UploadAsync(FileStorageReference reference, Stream stream)
@@ -49,9 +44,7 @@ public class DataLakeFileStorageClient : IFileStorageClient
         ArgumentNullException.ThrowIfNull(stream);
         ArgumentNullException.ThrowIfNull(reference);
 
-        var container = _clock.GetCurrentInstant() < _cutOffDate
-            ? _blobServiceClientObsoleted.GetBlobContainerClient(reference.Category.Value)
-            : _blobServiceClient.GetBlobContainerClient(reference.Category.Value);
+        var container = _blobServiceClient.GetBlobContainerClient(reference.Category.Value);
 
         stream.Position = 0; // Make sure we read the entire stream
         await container.UploadBlobAsync(reference.Path, stream).ConfigureAwait(false);
--- a/source/BuildingBlocks.Infrastructure/DataLakeFileStorageClient.cs
+++ b/source/BuildingBlocks.Infrastructure/DataLakeFileStorageClient.cs
@@ -18,6 +18,7 @@ using Azure.Storage.Blobs.Models;
 using Azure.Storage.Blobs.Specialized;
 using Energinet.DataHub.EDI.BuildingBlocks.Domain.Models;
 using Energinet.DataHub.EDI.BuildingBlocks.Infrastructure.Configuration.Options;
+using Energinet.DataHub.EDI.BuildingBlocks.Infrastructure.FeatureFlag;
 using Energinet.DataHub.EDI.BuildingBlocks.Interfaces;
 using Microsoft.Extensions.Azure;
 using Microsoft.Extensions.Options;
@@ -28,13 +29,16 @@ namespace Energinet.DataHub.EDI.BuildingBlocks.Infrastructure;
 
 public class DataLakeFileStorageClient : IFileStorageClient
 {
+    private readonly IFeatureFlagManager _featureFlagManager;
     private readonly BlobServiceClient _blobServiceClientObsoleted;
     private readonly BlobServiceClient _blobServiceClient;
 
     public DataLakeFileStorageClient(
         IAzureClientFactory<BlobServiceClient> clientFactory,
-        IOptions<BlobServiceClientConnectionOptions> options)
+        IOptions<BlobServiceClientConnectionOptions> options,
+        IFeatureFlagManager featureFlagManager)
     {
+        _featureFlagManager = featureFlagManager;
         _blobServiceClientObsoleted = clientFactory.CreateClient(options.Value.ClientNameObsoleted);
         _blobServiceClient = clientFactory.CreateClient(options.Value.ClientName);
     }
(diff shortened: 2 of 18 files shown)


source/B2CWebApi.AppTests/Fixture/B2CWebApiFixture.cs
```diff
@@ -126,6 +126,10 @@ public void SetTestOutputHelper(ITestOutputHelper? testOutputHelper)
                 $"{BlobServiceClientConnectionOptions.SectionName}:{nameof(BlobServiceClientConnectionOptions.StorageAccountUrl)}",
                 AzuriteManager.BlobStorageServiceUri.AbsoluteUri
             },
+            {
+                $"{BlobServiceClientConnectionOptions.SectionName}:{nameof(BlobServiceClientConnectionOptions.StorageAccountUrlObsoleted)}",
+                AzuriteManager.BlobStorageServiceUri.AbsoluteUri
```
lasrinnil, line 131: I take it that both clients are supposed to use `AzuriteManager.BlobStorageServiceUri.AbsoluteUri`

source/Tests/BuildingBlocks/FileStorageClient/DataLakeFileStorageClientTests.cs
```diff
@@ -0,0 +1,150 @@
...
+        await writer.FlushAsync();
+        stream.Position = 0;
+
+        // Act
+        await _sut.UploadAsync(reference, stream);
+
+        // Assert
+        _blobContainerClientMock.Verify(
+            x => x.UploadBlobAsync(reference.Path, stream, It.IsAny<CancellationToken>()),
+            Times.Once);
```
lasrinnil, line 103: Could we verify the blob is written to the "new" blob storage?

lasrinnil (dismissed):

source/BuildingBlocks.Infrastructure/FeatureFlag/FeatureFlagName.cs
```diff
@@ -49,4 +49,9 @@ public enum FeatureFlagName
     /// Whether to use orchestration for handling RequestAggregatedMeasureData processes.
     /// </summary>
     UseRequestAggregatedMeasureDataProcessOrchestration,
+
+    /// <summary>
+    /// Whether to start using standard blob service client.
```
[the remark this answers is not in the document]
MadsDue: It was what Outlaws referes to in regards of the new blob account. 

Hopefully, it will be removed again shortly

source/Tests/BuildingBlocks/FileStorageClient/DataLakeFileStorageClientTests.cs
MadsDue: Before merging this, I will deploy it to dev002 and verify it,

source/B2CWebApi.AppTests/Fixture/B2CWebApiFixture.cs
```diff
@@ -126,6 +126,10 @@ public void SetTestOutputHelper(ITestOutputHelper? testOutputHelper)
                 $"{BlobServiceClientConnectionOptions.SectionName}:{nameof(BlobServiceClientConnectionOptions.StorageAccountUrl)}",
                 AzuriteManager.BlobStorageServiceUri.AbsoluteUri
             },
+            {
+                $"{BlobServiceClientConnectionOptions.SectionName}:{nameof(BlobServiceClientConnectionOptions.StorageAccountUrlObsoleted)}",
```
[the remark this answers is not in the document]
MadsDue: Yes, that is the only one we have during integration test.

source/BuildingBlocks.Infrastructure/DataLakeFileStorageClient.cs
```diff
@@ -41,6 +49,10 @@ public async Task UploadAsync(FileStorageReference reference, Stream stream)
         ArgumentNullException.ThrowIfNull(reference);
 
         var container = _blobServiceClient.GetBlobContainerClient(reference.Category.Value);
+        if (!await _featureFlagManager.UseStandardBlobServiceClientAsync().ConfigureAwait(false))
```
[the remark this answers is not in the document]
MadsDue: I planned to remove it after everything was running on the new client.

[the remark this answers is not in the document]
ebbeknudsen: Yes, Maybe it makes sense to keep it for a little while (a few days after turning it on in production) so we can turn the feature off if we start seeing issue's.

source/BuildingBlocks.Infrastructure/FeatureFlag/FeatureFlagName.cs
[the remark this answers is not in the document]
ebbeknudsen: The correct name should be "classic" storage account, as far as i know that is what Azure calls it. However i don't really think it matters, because it will be deleted again shortly 👍

source/BuildingBlocks.Infrastructure/DataLakeFileStorageClient.cs
```diff
@@ -18,20 +18,28 @@
...
 public class DataLakeFileStorageClient : IFileStorageClient
 {
+    private readonly IFeatureFlagManager _featureFlagManager;
+    private readonly BlobServiceClient _blobServiceClientObsoleted;
     private readonly BlobServiceClient _blobServiceClient;
 
     public DataLakeFileStorageClient(
         IAzureClientFactory<BlobServiceClient> clientFactory,
-        IOptions<BlobServiceClientConnectionOptions> options)
+        IOptions<BlobServiceClientConnectionOptions> options,
```
ebbeknudsen, line 38: I'n not sure if we should do it now, but the client should definately be renamed to something like `BlobStorageFileStorageClient`, since it no longer uses the Data Lake storage account underneath.

source/Tests/BuildingBlocks/FileStorageClient/DataLakeFileStorageClientTests.cs
ebbeknudsen, line 103: We should probably also verify that `_blobContainerClientObsoletedMock` isn't called at all.

source/Tests/BuildingBlocks/FileStorageClient/DataLakeFileStorageClientTests.cs
```diff
@@ -0,0 +1,150 @@
...
+        _blobContainerClientObsoletedMock
+            .Setup(x => x.GetBlobClient(It.IsAny<string>()))
+            .Returns(_blobClientMock.Object);
+        // End: Obsoleted client
+
+        Mock<IFeatureFlagManager> featureFlagManager = new();
+        featureFlagManager
+            .Setup(x => x.UseStandardBlobServiceClientAsync())
+            .ReturnsAsync(true);
+        _sut = new DataLakeFileStorageClient(clientFactoryMock.Object, options, featureFlagManager.Object);
```
ebbeknudsen, line 82: Do we need a test that checks when the `UseStandardBlobServiceClientAsync()` feature flag returns false, the feature is not enabled? To ensure that production still uses the "obsolete" storage account, if the feature flag isn't turned on.

source/BuildingBlocks.Infrastructure/DataLakeFileStorageClient.cs
lasrinnil: Agreed. Lets move this to when we clean up after it has been running in prod for a bit.

source/Tests/BuildingBlocks/FileStorageClient/DataLakeFileStorageClientTests.cs
```diff
@@ -29,57 +29,53 @@ namespace Energinet.DataHub.EDI.Tests.BuildingBlocks.FileStorageClient;
...
+    [Fact]
+    public async Task Given_UploadAsync_When_UseStandardBlobServiceClientAsyncIsFalse_Then_ObsoleteBlobClientIsCalled()
+    {
+        // Arrange
+        var reference = new FileStorageReference(FileStorageCategory.ArchivedMessage(), "path");
+        var content = "test content";
+        var stream = new MemoryStream();
+        var writer = new StreamWriter(stream);
+        await writer.WriteAsync(content);
+        await writer.FlushAsync();
```
ebbeknudsen, line 58: Missing using / dispose of streams and writers

source/Tests/BuildingBlocks/FileStorageClient/DataLakeFileStorageClientTests.cs
```diff
@@ -0,0 +1,197 @@
...
+    [Fact]
+    public async Task Given_Stream_When_UploadAsync_Then_ShouldCallUploadBlobAsync()
+    {
+        // Arrange
+        var reference = new FileStorageReference(FileStorageCategory.ArchivedMessage(), "path");
+        var content = "test content";
+        var stream = new MemoryStream();
+        var writer = new StreamWriter(stream);
+        await writer.WriteAsync(content);
+        await writer.FlushAsync();
```
ebbeknudsen, line 90: Missing dispose

source/Tests/BuildingBlocks/FileStorageClient/DataLakeFileStorageClientTests.cs
```diff
@@ -0,0 +1,197 @@
...
+        stream.Position = 0;
+
+        Mock<IFeatureFlagManager> featureFlagManager = new();
+        featureFlagManager
+            .Setup(x => x.UseStandardBlobServiceClientAsync())
+            .ReturnsAsync(false);
+
+        var sut = new DataLakeFileStorageClient(GetClientFactoryMock(GetOptions()).Object, GetOptions(), featureFlagManager.Object);
+
+        // Act
```
ebbeknudsen, line 68: Wouldn't it be possible to just add the `featureFlagManager` to a private variable (`_featureFlagManager`) in the constructor, and then in this specific test overwrite `_featureFlagManager` to return false instead of true (using the `.Reset()` method? This way seems a little "hacky", where the `_sut` created in the constructor is only used in some tests, and the `GetClientFactoryMock` also overwrites the value of some private variables.

ebbeknudsen: (i can live with not changing this, since the tests are only meant to live a short amount of time 👍 )

ebbeknudsen (approved):

ebbeknudsen (approved):
LGTM

PR #1456, opened 2025-01-23
feat: Enqueue calculations from Process Manager

RasmusGraabaek:
## Checklist
- [ ] Reference to the task

### DIFF
--- a/source/BuildingBlocks.Infrastructure/FeatureFlag/FeatureFlagName.cs
+++ b/source/BuildingBlocks.Infrastructure/FeatureFlag/FeatureFlagName.cs
@@ -49,4 +49,10 @@ public enum FeatureFlagName
     /// Whether to use orchestration for handling RequestAggregatedMeasureData processes.
     /// </summary>
     UseRequestAggregatedMeasureDataProcessOrchestration,
+
+    /// <summary>
+    /// Whether to enqueue BRS-023/027 messages via the Process Manager.
+    /// Or to depend on the shared integration event topic
+    /// </summary>
+    EnqueueBrs023027MessagesViaProcessManager,
 }
--- a/source/BuildingBlocks.Infrastructure/FeatureFlag/IFeatureFlagManager.cs
+++ b/source/BuildingBlocks.Infrastructure/FeatureFlag/IFeatureFlagManager.cs
@@ -53,4 +53,10 @@ public interface IFeatureFlagManager
     /// Whether to use the RequestAggregatedMeasureData process orchestration.
     /// </summary>
     Task<bool> UseRequestAggregatedMeasureDataProcessOrchestrationAsync();
+
+    /// <summary>
+    /// Whether to enqueue BRS-023/027 messages via the Process Manager.
+    /// Or to depend on the shared integration event topic
+    /// </summary>
+    Task<bool> EnqueueBrs023027MessagesViaProcessManagerAsync();
 }
--- a/source/BuildingBlocks.Infrastructure/FeatureFlag/MicrosoftFeatureFlagManager.cs
+++ b/source/BuildingBlocks.Infrastructure/FeatureFlag/MicrosoftFeatureFlagManager.cs
@@ -40,5 +40,7 @@ public class MicrosoftFeatureFlagManager : IFeatureFlagManager
 
     public Task<bool> UseRequestAggregatedMeasureDataProcessOrchestrationAsync() => IsEnabledAsync(FeatureFlagName.UseRequestAggregatedMeasureDataProcessOrchestration);
 
+    public Task<bool> EnqueueBrs023027MessagesViaProcessManagerAsync() => IsEnabledAsync(FeatureFlagName.EnqueueBrs023027MessagesViaProcessManager);
+
     private Task<bool> IsEnabledAsync(FeatureFlagName featureFlagName) => _featureManager.IsEnabledAsync(featureFlagName.ToString());
 }
--- a/source/BuildingBlocks.Tests/TestDoubles/FeatureFlagManagerStub.cs
+++ b/source/BuildingBlocks.Tests/TestDoubles/FeatureFlagManagerStub.cs
@@ -34,4 +34,6 @@ public class FeatureFlagManagerStub : IFeatureFlagManager
     public Task<bool> UseRequestWholesaleServicesProcessOrchestrationAsync() => Task.FromResult(false);
 
     public Task<bool> UseRequestAggregatedMeasureDataProcessOrchestrationAsync() => Task.FromResult(false);
+
+    public Task<bool> EnqueueBrs023027MessagesViaProcessManagerAsync() => Task.FromResult(false);
 }


source/IntegrationEvents.IntegrationTests/Fixture/IntegrationEventsTestBase.cs
```diff
@@ -45,7 +51,6 @@ public void SetupServiceCollection()
             ["DB_CONNECTION_STRING"] = _integrationEventsFixture.DatabaseManager.ConnectionString,
             [$"{ServiceBusNamespaceOptions.SectionName}:{nameof(ServiceBusNamespaceOptions.FullyQualifiedNamespace)}"] = "Fake",
             [$"{BlobDeadLetterLoggerOptions.SectionName}:{nameof(BlobDeadLetterLoggerOptions.StorageAccountUrl)}"] = "https://fakeurl.com",
-            [$"{BlobDeadLetterLoggerOptions.SectionName}:{nameof(BlobDeadLetterLoggerOptions.ContainerName)}"] = "fake-container-name",
```
RasmusGraabaek, line 48: The tests are green, so is this blob stuff necessary?

source/B2BApi.AppTests/Functions/EnqueueMessages/BRS_023_027/EnqueueMessagesOrchestrationViaPMTests.cs
```diff
@@ -0,0 +1,620 @@
...
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Energinet.DataHub.EDI.B2BApi.AppTests.Functions.EnqueueMessages.BRS_023_027;
+
+[Collection(nameof(B2BApiAppCollectionFixture))]
```
RasmusGraabaek, line 44: This is pretty much a copy pasta of "EnqueueMessagesOrchestrationTests"

source/B2BApi.AppTests/Functions/EnqueueMessages/BRS_023_027/EnqueueMessagesOrchestrationTests.cs
```diff
@@ -58,6 +58,9 @@ public async Task InitializeAsync()
 
         // Clear mappings etc. before each test
         Fixture.ServiceBusListenerMock.ResetMessageHandlersAndReceivedMessages();
+        Fixture.EnsureAppHostUsesFeatureFlagValue(
+            enqueueBrs023027MessagesViaProcessManager: false,
+            disableEnqueueBrs023027MessagesFromWholesale: false);
```
ebbeknudsen, line 63: 1. Shouldn't one of these by `false` and the other one `true`?
2. Are these needed for all tests? Else it should be moved into the test that needs the app settings. Consider if it should be in the initial app settings in the fixture, since this adds a (maybe unnecessary) app restart.

source/B2BApi.AppTests/Functions/EnqueueMessages/BRS_023_027/EnqueueMessagesOrchestrationViaPMTests.cs
```diff
@@ -0,0 +1,620 @@
...
+public class EnqueueMessagesOrchestrationViaPMTests : IAsyncLifetime
```
ebbeknudsen, line 45:
```suggestion
public class EnqueueMessagesOrchestrationUsingProcessManagerTests : IAsyncLifetime
```

source/B2BApi.AppTests/Functions/EnqueueMessages/BRS_023_027/EnqueueMessagesOrchestrationViaPMTests.cs
ebbeknudsen: Alright, I won't review the test cases in depth then 👍

source/B2BApi.AppTests/Functions/EnqueueMessages/BRS_023_027/EnqueueMessagesOrchestrationViaPMTests.cs
```diff
@@ -0,0 +1,620 @@
...
+    }
+
+    /// <summary>
+    /// Verifies that:
+    ///  - The orchestration can complete a full run.
+    ///  - Every activity is executed once. We cannot be sure in which order, because we use fan-out/fan-in.
+    ///  - A service bus message is sent as expected.
+    /// </summary>
+    [Fact]
+    public async Task Given_CalculationOrchestrationId_When_CalculationCompletedEventForWholesaleFixingIsHandled_Then_OrchestrationCompletesWithExpectedServiceBusMessage()
```
ebbeknudsen, line 185: Should the naming of the tests all be renamed to `...When_EnqueueActorMessages...`, or do we keep it as-is to easily compare them with the tests they are copied from?

source/B2BApi.AppTests/Functions/EnqueueMessages/BRS_023_027/EnqueueMessagesOrchestrationViaPMTests.cs
ebbeknudsen: (i assume this is done because we delete the "EnqueueMessagesOrchestrationTests" in a couple of weeks)

source/B2BApi.AppTests/Functions/EnqueueMessages/BRS_023_027/EnqueueMessagesOrchestrationViaPMTests.cs
```diff
@@ -0,0 +1,620 @@
...
+        {
+            OrchestrationName = "Brs_023_027",
+            OrchestrationVersion = 1,
+            OrchestrationStartedByActorId = actorId,
+            OrchestrationInstanceId = orchestrationInstanceId?.ToString() ?? Guid.NewGuid().ToString(),
+        };
+        enqueueActorMessages.SetData(calculationCompletedEvent);
+        return enqueueActorMessages.ToServiceBusMessage(
+            subject: $"Enqueue_{enqueueActorMessages.OrchestrationName.ToLower()}",
+            idempotencyKey: "a-message-id");
```
ebbeknudsen, line 513: Shouldn't this idempotency key be a guid for EDI to work?

source/B2BApi/Functions/EnqueueMessages/BRS_023_027/Activities/SendActorMessagesEnqueuedActivity.cs
```diff
@@ -67,4 +80,30 @@ private static ServiceBusMessage CreateServiceBusMessage(ActorMessagesEnqueuedV1
...
+
+    private async Task SendToProcessManager(SendMessagesEnqueuedInput input)
+    {
+        await _processManagerMessageClient.NotifyOrchestrationInstanceAsync(
+                new NotifyOrchestrationInstanceEvent<NotifyEnqueueFinishedV1>(
+                    OrchestrationInstanceId: input.CalculationOrchestrationInstanceId,
+                    EventName: NotifyEnqueueFinishedV1.EventName,
+                    Data: new NotifyEnqueueFinishedV1 { Success = input.Success }),
+                CancellationToken.None)
+            .ConfigureAwait(false);
```
ebbeknudsen, line 107: I didn't think enqueue messages in EDI could fail. Don't we retry for unlimited time?

source/B2BApi/Functions/EnqueueMessages/BRS_023_027/Activities/SendActorMessagesEnqueuedActivity.cs
```diff
@@ -67,4 +80,30 @@ private static ServiceBusMessage CreateServiceBusMessage(ActorMessagesEnqueuedV1
...
+        await _wholesaleSender.SendMessageAsync(serviceBusMessage, CancellationToken.None).ConfigureAwait(false);
+    }
+
+    private async Task SendToProcessManager(SendMessagesEnqueuedInput input)
+    {
+        await _processManagerMessageClient.NotifyOrchestrationInstanceAsync(
+                new NotifyOrchestrationInstanceEvent<NotifyEnqueueFinishedV1>(
+                    OrchestrationInstanceId: input.CalculationOrchestrationInstanceId,
+                    EventName: NotifyEnqueueFinishedV1.EventName,
+                    Data: new NotifyEnqueueFinishedV1 { Success = input.Success }),
```
ebbeknudsen, line 105: Is `NotifyEnqueueFinishedV1` from the Process Manager NuGet package? Then it should be prefixed with something about the brs, like `CalculationActorMessagesEnqueuedNotifyEventV1`. I think this name is too broad if it is specific to the BRS 023/027.

source/B2BApi/Functions/EnqueueMessages/BRS_023_027/EnqueueHandler_Brs_023_027_V1.cs
```diff
@@ -12,30 +12,68 @@
...
+            "Received enqueue actor messages for BRS 023/027. Feature is {Status}. Data: {Data}",
+            featureIsDisabled ? "disabled" : "enabled",
             enqueueActorMessages.Data);
 
-        // TODO: Deserialize to actual input type instead of object (replace "object" type in summary as well)
-        var input = enqueueActorMessages.ParseData<object>();
+        if (featureIsDisabled)
+        {
+            await Task.CompletedTask.ConfigureAwait(false);
+        }
```
ebbeknudsen, line 54: Shouldn't this be possible?

```suggestion
        if (featureIsDisabled)
            return;
```

source/B2BApi/Functions/EnqueueMessages/BRS_023_027/EnqueueHandler_Brs_023_027_V1.cs
```diff
@@ -12,30 +12,68 @@
...
+
+        var calculationCompleted = enqueueActorMessages.ParseData<CalculatedDataForCalculationTypeV1>();
+        var orchestrationInput = new EnqueueMessagesOrchestrationInput(
+            CalculationId: calculationCompleted.CalculationId,
```
ebbeknudsen, line 58: You should probably guard against the data type being as expected.

Something like (taken from `EnqueueActorMessagesValidatedHandlerBase`):

```csharp
if (enqueueActorMessages.DataType == typeof(CalculatedDataForCalculationTypeV1).Name) {
	var data = enqueueActorMessages.ParseData<CalculatedDataForCalculationTypeV1>()
	HandleCalculatedDataForCalculationTypeV1(data);
}
else {
	throw exception
}
```

This makes it possible to handle/fail if/when CalculatedDataForCalculationTypeV1 version is bumped to V2, or if the wrong data type is sent.

source/B2BApi/Functions/EnqueueMessages/BRS_023_027/EnqueueHandler_Brs_023_027_V1.cs
```diff
@@ -12,30 +12,68 @@
...
+            CalculationOrchestrationId: enqueueActorMessages.OrchestrationInstanceId,
+            // This is not event driven anymore. But it is still passed to the outgoing messages
+            // Do we want to preserve this and use the orchestration instance id instead?
+            // enqueueActorMessages.OrchestrationInstanceId should be a guid, even tho the datatype states otherwise
+            EventId: Guid.Parse(enqueueActorMessages.OrchestrationInstanceId));
```
ebbeknudsen, line 63: I think we should use the message id from the service bus message instead of the orchestration instance id, for this to make sense.

source/B2BApi/Functions/EnqueueMessages/BRS_023_027/EnqueueHandler_Brs_023_027_V1.cs
```diff
@@ -12,30 +12,68 @@
...
+        var instanceId = await durableClient.StartNewAsync(
+            nameof(EnqueueMessagesOrchestration),
+            orchestrationInput).ConfigureAwait(false);
+
+        _logger.LogInformation(
+            "Started 'EnqueueMessagesOrchestration' (id '{OrchestrationInstanceId}', calculation id: {CalculationId}, calculation type: {CalculationType}, calculation orchestration id: {CalculationOrchestrationId}.",
+            instanceId,
+            calculationCompleted.CalculationId,
+            calculationCompleted.CalculationType,
+            enqueueActorMessages.OrchestrationInstanceId);
```
ebbeknudsen, line 76: Maybe add the message id from the service bus message as well.

source/B2BApi/Functions/EnqueueMessages/BRS_023_027/EnqueueHandler_Brs_023_027_V1.cs
```diff
@@ -12,30 +12,68 @@
...
 
         // TODO: Call actual logic that enqueues messages (starts orchestration)
         await Task.CompletedTask.ConfigureAwait(false);
```
ebbeknudsen, line 79: This can be removed now

source/BuildingBlocks.Infrastructure/FeatureFlag/FeatureFlagName.cs
```diff
@@ -54,4 +54,14 @@ public enum FeatureFlagName
...
+
+    /// <summary>
+    /// Whether to enqueue BRS-023/027 messages via the Process Manager.
+    /// </summary>
+    EnqueueBrs023027MessagesFromProcessManager,
+
+    /// <summary>
+    /// Whether to disable enqueue BRS-023/027 messages via Wholesale.
+    /// </summary>
+    DisableEnqueueBrs023027MessagesFromWholesale,
```
ebbeknudsen, line 66: Consider if these makes sense.

1. Can't we just use the `UseRequestWholesaleServicesProcessOrchestration` and `UseRequestAggregatedMeasureDataProcessOrchestration` feature flags? Does it ever make sense to only enable parts of the Process Manager flow? I think we must either enable the entire process manager flow, or nothing at all, for EDI to work.
2. **IF** we keep the `EnqueueBrs023027MessagesFromProcessManager`, i don't think it makes sense to have a feature flag for `DisableEnqueueBrs023027MessagesFromWholesale` as well. I don't think EDI will work of both `EnqueueBrs023027MessagesFromProcessManager` is true but `DisableEnqueueBrs023027MessagesFromWholesale` is false. Enqueue from Wholesale should just be disabled if `EnqueueBrs023027MessagesFromProcessManager` is true.

source/BuildingBlocks.Infrastructure/FeatureFlag/FeatureFlagName.cs
```diff
@@ -54,4 +54,14 @@ public enum FeatureFlagName
     /// Whether to start using standard blob service client.
     /// </summary>
     UseStandardBlobServiceClient,
+
+    /// <summary>
+    /// Whether to enqueue BRS-023/027 messages via the Process Manager.
+    /// </summary>
+    EnqueueBrs023027MessagesFromProcessManager,
```
ebbeknudsen, line 61: If kept, follow the naming convention of the other feature flags:

```suggestion
    UseEnqueueBrs023027MessagesFromProcessManager,
```

source/BuildingBlocks.Tests/TestDoubles/DurableClientFactoryStub.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextImplementations;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
+
+namespace Energinet.DataHub.EDI.BuildingBlocks.Tests.TestDoubles;
+
+public class DurableClientFactoryStub : IDurableClientFactory
+{
+    public IDurableClient CreateClient(DurableClientOptions durableClientOptions)
+    {
+        throw new Exception("This method is just here to test feature flags");
```
ebbeknudsen, line 25:
```suggestion
        throw new NotImplementedException("This method is just here to test feature flags");
```

source/BuildingBlocks.Tests/TestDoubles/DurableClientFactoryStub.cs
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public IDurableClient CreateClient()
+    {
+        throw new Exception("This method is just here to test feature flags");
```
ebbeknudsen, line 30:
```suggestion
        throw new NotImplementedException("This method is just here to test feature flags");
```

source/IntegrationEvents.Application/Extensions/DependencyInjection/IntegrationEventExtensions.cs
```diff
@@ -17,6 +17,7 @@
 using Energinet.DataHub.Core.Messaging.Communication.Extensions.DependencyInjection;
 using Energinet.DataHub.Core.Messaging.Communication.Extensions.Options;
 using Energinet.DataHub.Core.Messaging.Communication.Subscriber;
+using Energinet.DataHub.EDI.BuildingBlocks.Infrastructure.Extensions.DependencyInjection;
```
ebbeknudsen, line 20: Didn't we reverse the dependencies, to follow clean architecture? So the references must follow this: `Infrastructure -> Application -> Domain`, so `IntegrationEvents.Application` is not allowed to reference `BuildingBlocks.Infrastructure`

source/B2BApi.AppTests/Functions/EnqueueMessages/BRS_023_027/EnqueueMessagesOrchestrationViaPMTests.cs
RasmusGraabaek: I was planning on keep changing as little as possible

source/IntegrationEvents.Infrastructure/EventProcessors/CalculationCompletedV1Processor.cs
```diff
@@ -24,19 +25,27 @@ public sealed class CalculationCompletedV1Processor : IIntegrationEventProcessor
...
         _durableClientFactory = durableClientFactory;
+        _featureFlagManager = featureFlagManager;
     }
 
     public string EventTypeToHandle => CalculationCompletedV1.EventName;
 
     public async Task ProcessAsync(IntegrationEvent integrationEvent, CancellationToken cancellationToken)
     {
+        if (await _featureFlagManager.DisableEnqueueBrs023027MessagesFromWholesaleAsync().ConfigureAwait(false))
+        {
```
ebbeknudsen, line 45:
```suggestion
	var useEnqueueFromWholesale = _featureFlagManager.UseEnqueueBrs023027MessagesFromWholesaleAsync().ConfigureAwait(false);
        if (!useEnqueueFromWholesale)
        {
```

source/IntegrationEvents.Infrastructure/Model/EnqueueMessagesOrchestrationInput.cs
```diff
@@ -12,14 +12,15 @@
...
 /// <summary>
-/// An immutable input to start the equeue messages orchestration.
+/// An immutable input to start the enqueue messages orchestration.
 /// </summary>
 public sealed record EnqueueMessagesOrchestrationInput(
     string CalculationOrchestrationId,
     Guid CalculationId,
+    // This is not always an eventId, but may also be a orchestrationId.
+    // Should it be renamed to orchestrationId when the integration event enqueuer is deleted?
+    // Maybe ExternalId?
```
ebbeknudsen, line 25: I think this should still be an event id, since the orchestration is still started from a service bus event, it just originates from the Process Manager instead of from Wholesale

source/IntegrationEvents.IntegrationTests/Fixture/IntegrationEventsTestBase.cs
ebbeknudsen: If the tests still work i don't have a problem in deleting it.

source/IntegrationEvents.IntegrationTests/Tests/WhenCalculatedCompletedEventIsReceived.cs
```diff
@@ -0,0 +1,89 @@
...
+        };
+
+        var act = () => HavingReceivedAndHandledIntegrationEventAsync(integrationEvent);
+
+        await act.Should().NotThrowAsync("the handler should not do anything");
+    }
+
+    private async Task HavingReceivedAndHandledIntegrationEventAsync(CalculationCompletedV1 calculationCompleted)
+    {
+        var integrationEventHandler = Services.GetService<IIntegrationEventHandler>();
```
ebbeknudsen, line 79:
```suggestion
        var integrationEventHandler = Services.GetRequiredService<IIntegrationEventHandler>();
```

source/IntegrationEvents.IntegrationTests/Tests/WhenCalculatedCompletedEventIsReceived.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        var integrationEvent = new IntegrationEvent(
+            Guid.NewGuid(),
+            CalculationCompletedV1.EventName,
+            1,
+            calculationCompleted);
+
+        await integrationEventHandler!.HandleAsync(integrationEvent).ConfigureAwait(false);
```
ebbeknudsen, line 87:
```suggestion
        await integrationEventHandler.HandleAsync(integrationEvent).ConfigureAwait(false);
```

source/B2BApi/Functions/EnqueueMessages/BRS_023_027/Activities/SendActorMessagesEnqueuedActivity.cs
```diff
@@ -25,33 +29,42 @@ namespace Energinet.DataHub.EDI.B2BApi.Functions.EnqueueMessages.BRS_023_027.Act
...
         [ActivityTrigger] SendMessagesEnqueuedInput input)
     {
-        var messagesEnqueuedEvent = new ActorMessagesEnqueuedV1
-        {
-            OrchestrationInstanceId = input.CalculationOrchestrationInstanceId,
-            CalculationId = input.CalculationId.ToString(),
-            Success = input.Success,
-        };
+        // By construction, we have that the process manager orchestration has the same id as the calculation
+        var instructionFromProcessManager = input.CalculationId.ToString() == input.CalculationOrchestrationInstanceId;
```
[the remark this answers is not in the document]
RasmusGraabaek: I'm unsure on this aswell.
But the argument for this is, if we have enqueues from PM active and Wholesale simultaneously. Then we would get an exception in either PM or wholesale, if we sent a notify to both of them.

But the brings the question of: Why two flags?

This should make it easier to make the switch. Since we do not have to think of running enqueuements

source/B2BApi/Functions/EnqueueMessages/BRS_023_027/Activities/SendActorMessagesEnqueuedActivity.cs
RasmusGraabaek: "input.Success" ?

If no messages are enqueued but we reached this activity, then we say that it failed.

source/IntegrationEvents.IntegrationTests/Tests/WhenCalculatedCompletedEventIsReceived.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            CalculationId = Guid.NewGuid().ToString(),
+            CalculationVersion = 1,
+            InstanceId = "instance",
+            CalculationType = CalculationCompletedV1.Types.CalculationType.Aggregation,
+        };
+
+        var act = () => HavingReceivedAndHandledIntegrationEventAsync(integrationEvent);
+
+        await act.Should().NotThrowAsync("the handler should not do anything");
```
ebbeknudsen, line 74: I think this is pretty "hacky". I suggest setting a variable in the `DurableClientFactoryStub` if the method is called, and then assert if the variable is `true`/`false`. We do something similar in the `ServiceBusSenderSpy`.

The `DurableClientFactoryStub` should probably be renamed to `DurableClientFactorySpy` as well.

source/IntegrationEvents.Application/Extensions/DependencyInjection/IntegrationEventExtensions.cs
ebbeknudsen: I see the dependencies havn't been reversed for the `IntegrationEvents` projects, so you can ignore this comment 🥲

source/IntegrationEvents.IntegrationTests/Tests/WhenCalculatedCompletedEventIsReceived.cs
```diff
@@ -0,0 +1,89 @@
...
+            CalculationVersion = 1,
+            InstanceId = "instance",
+            CalculationType = CalculationCompletedV1.Types.CalculationType.Aggregation,
+        };
+
+        var act = () => HavingReceivedAndHandledIntegrationEventAsync(integrationEvent);
+
+        await act.Should().ThrowAsync<Exception>("the handler should try to create a durable client,"
+                                                 + "which is mocked to throw an exception")
+            .WithMessage("This method is just here to test feature flags");
```
ebbeknudsen, line 56: I think this is pretty "hacky". I suggest setting a variable in the DurableClientFactoryStub if the method is called, and then assert if the variable is true/false. We do something similar in the ServiceBusSenderSpy.

The DurableClientFactoryStub should probably be renamed to DurableClientFactorySpy as well.

source/IntegrationEvents.IntegrationTests/Tests/WhenCalculatedCompletedEventIsReceived.cs
ebbeknudsen, line 74:
```suggestion
        await act.Should().NotThrowAsync("the durable client factory should not be called since the feature is disabled");
```

source/B2BApi.AppTests/Functions/EnqueueMessages/BRS_023_027/EnqueueMessagesOrchestrationViaPMTests.cs
ebbeknudsen: Good idea, we should probably only rename them after deleting the old ones 👍

source/B2BApi/Functions/EnqueueMessages/BRS_023_027/EnqueueHandler_Brs_023_027_V1.cs
RasmusGraabaek: The messageId is just a string. And I'm unsure on this is really true, Or that it's a guid in disguise

source/B2BApi/Functions/EnqueueMessages/BRS_023_027/Activities/SendActorMessagesEnqueuedActivity.cs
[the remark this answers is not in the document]
ebbeknudsen: We should probably discuss this with others in the team, maybe at daily scrum tomorrow. I think it would be far less complicated to not have any running orchestrations when we make the switch, and we would need to handle far fever edge cases. And i think it will be easy to find a time where there is no orchestration running. We could do something like:
1. Disabling starting new orchestrations in Wholesale
1. Wait for orchestrations to complete (if any)
1. Release EDI/PM feature

source/IntegrationEvents.Application/Extensions/DependencyInjection/IntegrationEventExtensions.cs
RasmusGraabaek: Not for integration events it seems.
It's the only module which has the dependencyInjection stuff in application.

I can fix it in another pull request.

source/IntegrationEvents.IntegrationTests/Tests/WhenCalculatedCompletedEventIsReceived.cs
RasmusGraabaek: I have added a DurableClient - spy
Such that the factory returns the spy and we may assert on the number of jobs started by the client.

source/BuildingBlocks.Tests/TestDoubles/DurableClientSpy.cs
```diff
@@ -0,0 +1,244 @@
...
+    private readonly string _taskHubName1 = "taskHubName2";
+    private readonly string _taskHubName2 = "taskHubName3";
+
+    string IDurableClient.TaskHubName => _taskHubName2;
+
+    string IDurableEntityClient.TaskHubName => _taskHubName1;
+
+    string IDurableOrchestrationClient.TaskHubName => _taskHubName;
+
+    public int NumberOfJobsStarted { get; private set; } = 0;
```
RasmusGraabaek, line 34: This is the important variable.
Since it will increase by 1 everytime someone calls "StartNewAsync"
Everything else throws a NotImplementedException

source/B2BApi/Functions/EnqueueMessages/EnqueueActorMessagesHandlerBase.cs
```diff
@@ -64,7 +62,7 @@ public async Task EnqueueAsync(ServiceBusReceivedMessage message)
         }
     }
 
-    protected abstract Task EnqueueActorMessagesV1Async(EnqueueActorMessagesV1 enqueueActorMessages);
+    protected abstract Task EnqueueActorMessagesV1Async(EnqueueActorMessagesV1 enqueueActorMessages, string eventId);
```
ebbeknudsen, line 65: I would prefer `eventId` to be called `serviceBusMessageId`, but i can live with this if you don't want to change it ;)

source/IntegrationEvents.IntegrationTests/Fixture/IntegrationEventsTestBase.cs
```diff
@@ -61,6 +62,7 @@ public void SetupServiceCollection()
             .AddMasterDataModule(configuration)
             .AddIntegrationEventModule(configuration)
             .AddTransient<IFeatureFlagManager>(_ => FeatureFlagManagerStub)
+            .AddScoped<IDurableClient, DurableClientSpy>()
             .AddScoped<IDurableClientFactory, DurableClientFactoryStub>();
```
ebbeknudsen: You could add the `DurableClientSpy` as a protected/public property, so you don't have to get it (and cast it, with null check) through services when you need it in tests

source/IntegrationEvents.IntegrationTests/Tests/WhenCalculatedCompletedEventIsReceived.cs
```diff
@@ -49,17 +50,16 @@ public async Task Given_CalculationCompletedV1_When_FeatureIsEnabled_Then_EventH
...
         };
 
-        var act = () => HavingReceivedAndHandledIntegrationEventAsync(integrationEvent);
+        await HavingReceivedAndHandledIntegrationEventAsync(integrationEvent);
 
-        await act.Should().ThrowAsync<Exception>("the handler should try to create a durable client,"
-                                                 + "which is mocked to throw an exception")
-            .WithMessage("This method is just here to test feature flags");
+        var durableClientStub = Services.GetRequiredService<IDurableClient>() as DurableClientSpy;
+        durableClientStub!.NumberOfJobsStarted.Should().Be(1);
```
ebbeknudsen, line 56: You could add the `DurableClientSpy` as a protected/public property, so you don't have to get it (and cast it, with null check) through services when you need it in tests

source/IntegrationEvents.IntegrationTests/Tests/WhenCalculatedCompletedEventIsReceived.cs
```diff
@@ -69,21 +69,22 @@ public async Task Given_CalculationCompletedV1_When_FeatureIsDisabled_Then_Event
             CalculationType = CalculationCompletedV1.Types.CalculationType.Aggregation,
         };
 
-        var act = () => HavingReceivedAndHandledIntegrationEventAsync(integrationEvent);
+        await HavingReceivedAndHandledIntegrationEventAsync(integrationEvent);
 
-        await act.Should().NotThrowAsync("the handler should not do anything");
+        var durableClientStub = Services.GetRequiredService<IDurableClient>() as DurableClientSpy;
+        durableClientStub!.NumberOfJobsStarted.Should().Be(0);
```
ebbeknudsen, line 75: You could add the `DurableClientSpy` as a protected/public property, so you don't have to get it (and cast it, with null check) through services when you need it in tests

ebbeknudsen (dismissed):

ebbeknudsen (approved):
LGTM

djorgensendk (approved):

PR #1464, opened 2025-01-31
feat: log if DataLakeFileStorageClient uses obsolete blobclient

lasrinnil:
## Description
This logging is to help ensure that the migration of the obsolete document storage is succesful, will be removed when we have verified the migration and the `DataLakeFileStorageClient` is cleaned up.

## Checklist
- [ ] Reference to the task

### DIFF
--- a/source/BuildingBlocks.Infrastructure/DataLakeFileStorageClient.cs
+++ b/source/BuildingBlocks.Infrastructure/DataLakeFileStorageClient.cs
@@ -89,7 +89,7 @@ public class DataLakeFileStorageClient : IFileStorageClient
         if (blobExists == null || blobExists is { Value: false })
         {
             // This logging is only temporary while we verify data migration is successful.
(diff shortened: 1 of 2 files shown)


MWO1024 (approved):

source/BuildingBlocks.Infrastructure/DataLakeFileStorageClient.cs
```diff
@@ -84,6 +88,8 @@ public async Task<FileStorageFile> DownloadAsync(FileStorageReference reference,
         var blobExists = await blob.ExistsAsync(cancellationToken).ConfigureAwait(false);
         if (blobExists == null || blobExists is { Value: false })
         {
+            // This logging is only temporary while we verify data migration is successful.
+            _logger.LogInformation($"Blob does not exist in the new storage account, trying the obsoleted storage account. Category {reference.Category.Value}, Path {reference.Path}");
```
ebbeknudsen, line 92: We should use log arguments (structured logging) instead of string concatonation.

ebbeknudsen (approved):

ebbeknudsen (approved):

PR #1472, opened 2025-02-05
refac: Rename EffectSettlement to CapacitySettlement

AndersBallingPetersen:
## Description
Rename EffectSettlement to CapacitySettlement.

## Checklist
- [ ] Reference to the task


source/OutgoingMessages.Interfaces/Models/CalculationResults/MeteringPointType.cs
```diff
@@ -33,5 +33,5 @@ public enum MeteringPointType
     TotalConsumption = 11,
     ElectricalHeating = 12,
     NetConsumption = 13,
-    EffectSettlement = 14,
```
RasmusGraabaek, line 36: Is this serialized in blob?

source/Contracts/Requests/WholesaleServices/WholesaleServicesRequestAccepted.proto
```diff
@@ -111,6 +111,7 @@ message WholesaleServicesRequestSeries {
     METERING_POINT_TYPE_ELECTRICAL_HEATING = 12;
     METERING_POINT_TYPE_NET_CONSUMPTION = 13;
     METERING_POINT_TYPE_EFFECT_SETTLEMENT = 14;
+    METERING_POINT_TYPE_CAPACITY_SETTLEMENT = 15;
```
MadsDue, line 114: Do you think it's necessary to update this now? Will it not be removed when we no longer use Wholesale but PM.

However, the Wholesale repo needs to be updated to match this change as well.

source/OutgoingMessages.Interfaces/Models/CalculationResults/MeteringPointType.cs
MadsDue: I am not sure what this type is. Is this used in context of PM, when PM is sending the enqueue event?

PR #1486, opened 2025-02-11
test: wholesale delegation via pm behaviour tests

MadsDue:
## Description

Copied all tests from GivenWholesaleServicesRequestWithDelegationTests into GivenWholesaleServicesRequestV2WithDelegationTests and refactored to use Process Manager instead of the old Process Module and wholesale.

## Checklist
- [ ] Reference to the task

### DIFF
--- a/source/IntegrationTests/Behaviours/IncomingRequests/GivenWholesaleServicesRequestV2WithDelegationTests.cs
+++ b/source/IntegrationTests/Behaviours/IncomingRequests/GivenWholesaleServicesRequestV2WithDelegationTests.cs
@@ -511,7 +511,7 @@ public class GivenWholesaleServicesRequestV2WithDelegationTests : WholesaleServi
         var expectedErrorCode = "E17";
 
         // Generate a mock ServiceBus Message with RequestCalculatedWholesaleServicesAcceptedV1 response from Process Manager,
-        // based on the RequestCalculatedWholesaleServicesInputV1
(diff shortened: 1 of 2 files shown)


source/IntegrationTests/Behaviours/IncomingRequests/GivenWholesaleServicesRequestV2WithDelegationTests.cs
```diff
@@ -0,0 +1,1064 @@
...
+                ChargeCode: expectedChargeCode,
+                ChargeType: expectedChargeType,
+                Currency: exampleWholesaleResultMessageForActor.Currency,
+                EnergySupplierNumber: energySupplierNumber.Value,
+                SettlementMethod: exampleWholesaleResultMessageForActor.SettlementMethod,
+                MeteringPointType: exampleWholesaleResultMessageForActor.MeteringPointType,
+                GridArea: gridAreaCode,
+                transactionId,
+                PriceMeasurementUnit: exampleWholesaleResultMessageForActor.MeasurementUnit,
+                ProductCode: "5790001330590", // Example says "8716867000030", but document writes as "5790001330590"?
```
RasmusGraabaek, line 249:
```suggestion
                ProductCode: "5790001330590", // Example says "8716867000030", but document writes as "5790001330590"?
```
If the comment is copied. I think we should delete it

RasmusGraabaek (approved):

PR #1504, opened 2025-02-18
test: added test of amount type mapper + clean up

MadsDue:
## Description
Added a new AmountType Mapper unit test and removed old wholesale AmountType Mapper + test.

Introduced new folders:
 source/Tests/B2BApi/Functions/EnqueueMessages/Mappers/
 - AmountTypeMapperTests.cs
source/B2BApi/Functions/EnqueueMessages/Mappers/
 - AmountTypeMapper.cs

## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/13391414195
- [x] Reference to the task


source/Edi/Factories/WholesaleServicesRequestMapper.cs
```diff
@@ -35,7 +36,8 @@ public IReadOnlyCollection<WholesaleServicesRequest> Map(Energinet.DataHub.Edi.R
         var resolution = request.HasResolution ? ResolutionMapper.Map(request.Resolution) : (Resolution?)null;
 
         // If no charge types are requested, both monthly amount and total monthly amount is requested
-        var amountTypes = AmountTypeMapper.Map(resolution, AllChargesIsRequested(request));
+        // NOT USED!
+        var amountTypes = new List<AmountType>();
```
lasrinnil, line 40: Should this not be removed if the collection is never used?

lasrinnil (approved):

source/Edi/Factories/WholesaleServicesRequestMapper.cs
MadsDue: It is part of what has been copied from wholesale as I understand it and is not used. 

This was a quick fix to make the compiler happy until the whole thing is removed.

lasrinnil (approved):

PR #1506, opened 2025-02-19
fix: Use PM event in load test

MWO1024:
## Description

Since t001 has been updated to use the PM version of the calculation flow, EDI will no longer act upon the event from wholesale. This had the side effect of making the load test do nothing, as it was based on reception of an event from wholesale.

The load test has been updated to instead publish the PM event. While we could have looked at the feature flag and used that to determine which event to publish, we are so close to fully changing to the PM flow that it would be a wasted effort.

## References

- d002 subsystem tests: https://github.com/Energinet-DataHub/dh3-environments/actions/runs/13414132472
- d002 load tests: https://github.com/Energinet-DataHub/dh3-environments/actions/runs/13412438944

## Checklist
- [ ] Reference to the task


source/SubsystemTests/Dsl/CalculationCompletedDsl.cs
```diff
@@ -63,9 +62,7 @@ internal static async Task<DurableOrchestrationStatus> StartEnqueueMessagesOrche
         var orchestrationStartedAfter = SystemClock.Instance.GetCurrentInstant().Minus(Duration.FromSeconds(10));
 
         logger.WriteLine("Publish calculation completed for calculation with id {0}", calculationId);
```
RasmusGraabaek, line 64:
```suggestion
        logger.WriteLine("Publish nameof(EnqueueActorMessagesV1) with nameof(CalculationEnqueueActorMessagesV1) for calculation with id {0}", calculationId);
```

source/SubsystemTests/Dsl/CalculationCompletedDsl.cs
```diff
@@ -52,9 +52,8 @@ internal CalculationCompletedDsl(EdiDriver ediDriver, EdiDatabaseDriver ediDatab
 
     internal static async Task<DurableOrchestrationStatus> StartEnqueueMessagesOrchestration(
         ITestOutputHelper logger,
-        WholesaleDriver wholesaleDriver,
+        ProcessManagerDriver processManagerDriver,
```
RasmusGraabaek, line 55: I may be wrong. but I think the subsystem will fail, since this is used in PublishForBalanceFixingCalculation which is currently being run

MWO1024: It did indeed. I'll figure something out.

PR #1510, opened 2025-02-20
feat: Remove PM from EDI

MWO1024:
## References

https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/388

- d002 deploy: https://github.com/Energinet-DataHub/dh3-environments/actions/runs/13451886792

## Checklist
- [x] Reference to the task


dstenroejl (commented):
We should also delete MD files:
image

dstenroejl (commented):
I have verified that our packages on NuGet gallery are using the same names, so we DO NOT need to deprecate them (unless we like to reduce numbers of versions shown to developers). :)

dstenroejl (commented):
Shared files should also be deleted
image

.github/workflows/detect-changes.yml
dstenroejl, line 1: Search this file for 'processmanager_client' and you will find more to delete

dstenroejl (commented):
One last comment regarding a YML file :)

dstenroejl (approved):
LGTM

NicolajAaH (approved):

PR #1524, opened 2025-03-06
test: Implement BRS021 enqueue message activity behavior test

AndersBallingPetersen:
## Checklist
- [ ] Reference to the task


source/IntegrationTests/Behaviours/IncomingRequests/GivenMeteredDataForMeteringPointV2Tests.cs
```diff
@@ -0,0 +1,187 @@
...
+    [MemberData(nameof(SupportedDocumentFormats))]
+    public async Task When_ActorPeeksMessage_Then_ReceivesOneDocumentWithCorrectContent(DocumentFormat documentFormat)
+    {
+        /*
+         *  --- PART 1: Receive request and send message to Process Manager ---
+         */
+
+        // Arrange
+        var senderSpy = CreateServiceBusSenderSpy(ServiceBusSenderNames.ProcessManagerStartSender);
+        var senderActor = (ActorNumber: ActorNumber.Create("1111111111111"), ActorRole: ActorRole.GridAccessProvider);
```
MWO1024, line 54: Depending on which get PR gets out first, we might have to use the new BRS-021 topic/subscriptions here.

source/IntegrationTests/EventBuilders/MeteredDataForMeteringPointEventBuilder.cs
```diff
@@ -0,0 +1,110 @@
...
+        var acceptRequest = new ForwardMeteredDataAcceptedV1(
+            OriginalActorMessageId: requestMeteredDataForMeteringPointMessageInputV1.MessageId,
+            MeteringPointId: requestMeteredDataForMeteringPointMessageInputV1.MeteringPointId,
+            MeteringPointType: meteringPointType,
+            OriginalTransactionId: requestMeteredDataForMeteringPointMessageInputV1.TransactionId,
+            ProductNumber: requestMeteredDataForMeteringPointMessageInputV1.ProductNumber,
+            MeasureUnit: measureUnit,
+            RegistrationDateTime: InstantPattern.General.Parse(requestMeteredDataForMeteringPointMessageInputV1.RegistrationDateTime).Value.ToDateTimeOffset(),
+            Resolution: resolution,
+            StartDateTime: invariantPattern.Parse(requestMeteredDataForMeteringPointMessageInputV1.StartDateTime).Value.ToDateTimeOffset(),
```
MWO1024, line 86: Is there a reason these two are not part of the parsing and validation above?

source/Process.Application/ProcessInitializationHandlers/InitializeMeteredDataForMeteringPointHandler.cs
```diff
@@ -65,7 +65,7 @@ await _outgoingMessagesClient.EnqueueAndCommitAsync(
                         new MeteredDataForMeteringPointMessageSeriesDto(
                             TransactionId.From(string.Join(string.Empty, series.TransactionId.Reverse())),
                             series.MeteringPointLocationId!,
-                            series.MeteringPointType!,
+                            MeteringPointType.FromCode(series.MeteringPointType!),
                             TransactionId.From(series.TransactionId),
                             series.ProductNumber!,
```
MWO1024, line 70: I think we should be able to delete this now, since we are using the PM. I'll add a note on my todo-list, so let us do it in another PR.

source/OutgoingMessages.IntegrationTests/DocumentAsserters/NotifyValidatedMeasureDataDocumentAsserter.cs
```diff
@@ -79,7 +79,7 @@ public static async Task AssertCorrectDocumentAsync(
 
             // Required series fields
             asserter
-                .HasTransactionId(seriesIndex, requiredSeriesFields.TransactionId)
+                .TransactionIdExists(seriesIndex)
                 .HasMeteringPointNumber(
```
MWO1024, line 83: Are we no longer in control over the transaction id?

MWO1024 (changes requested):
Added a few comments.

source/OutgoingMessages.IntegrationTests/DocumentAsserters/NotifyValidatedMeasureDataDocumentAsserter.cs
AndersBallingPetersen: The ProcessManager always creates a new guid. We don't know the value.

MWO1024 (approved):

PR #1529, opened 2025-03-10
refac: Enable Forward Metered Data in EBIX format

dstenroejl:
## References

https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/587

## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/13784723170
- [x] Reference to the task


source/IncomingMessages.Domain/Messages/MeteredDataForMeteringPointMessage.cs
```diff
@@ -71,7 +71,6 @@ public record MeteredDataForMeteringPointSeries(
     string StartDateTime,
     string? EndDateTime,
     string? ProductNumber,
-    // This field is not used in Ebix
     string? RegisteredAt,
```
MadsDue, line 74: Can't we remove nullability in this field?

dstenroejl: I didn't do it because of how the CIM XML parser is implemented. But if we are OK with that change, I can do it. You can just look at the latest commit and decide.

MadsDue (approved):

PR #1552, opened 2025-03-26
refac: B2C RequestWholesaleSettlementTemp

lasrinnil:
## Description
We have decided that client logic should be contained in the BFF, so this is the first step in exposing a simpler model for creating wholesalesettlement requests. Without the client logic imposed on the model `RequestWholesaleServicesMarketDocumentV2`

- This "Temp" approch is used since RequestWholesaleSettlementController did not support versioning.
- After the BFF is moved to the "Temp" endpoint. RequestWholesaleSettlementController will be refactored to expose same endpoint as RequestWholesaleSettlementControllerTemp did, but now with versioning.

- Updated nswag to target donet 8
- Updated all swagger's after nswag uses dotnet 8
## References
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/548

## Checklist
- [ ] Reference to the task


source/B2CWebApi/Controllers/TempRequestWholesaleSettlementController.cs
```diff
@@ -0,0 +1,138 @@
...
+using ActorRole = Energinet.DataHub.EDI.BuildingBlocks.Domain.Models.ActorRole;
+
+namespace Energinet.DataHub.EDI.B2CWebApi.Controllers;
+
+[ApiController]
+[Route("[controller]")]
+public class TempRequestWholesaleSettlementController : ControllerBase
+{
+    private const string WholesaleSettlementMessageType = "D21";
+    private const string WholesaleSettlementBusinessType = "23";
```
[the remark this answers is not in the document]
RasmusGraabaek: One may argue that the name "BusinessType" is a bit vague

source/B2CWebApi/Controllers/TempRequestWholesaleSettlementController.cs
```diff
@@ -0,0 +1,138 @@
...
+                activity: AuditLogActivity.RequestWholesaleResults,
+                activityOrigin: HttpContext.Request.GetDisplayUrl(),
+                activityPayload: request,
+                affectedEntityType: AuditLogEntityType.RequestWholesaleServices,
+                affectedEntityKey: null)
+            .ConfigureAwait(false);
+
+        var currentUser = _userContext.CurrentUser;
+
+        var message = new RequestWholesaleSettlementDto(
```
RasmusGraabaek, line 79: Is there an arguement for not using "RequestWholesaleSettlementDtoFactory" ?

RasmusGraabaek: It contains logic which ensures that the given start and end time is at midnight.

The last time I looking we very given something which wasn't at mightnigt but seconds away from it

eg. 2025-3-28T22:59:00 or something like that

source/B2CWebApi/Controllers/TempRequestWholesaleSettlementController.cs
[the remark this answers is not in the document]
RasmusGraabaek: With this in mind. We should problably just hardcode it into the DTO

source/SubsystemTests/Drivers/B2C/B2CEdiDriver.cs
```diff
@@ -94,6 +97,44 @@ await webApiClient.RequestWholesaleSettlementAsync(
...
+            {
+                new RequestWholesaleSettlementSeriesV2
+                {
+                    Id = Guid.NewGuid().ToString(),
+                    StartDateAndOrTimeDateTime = start.ToString(),
+                    EndDateAndOrTimeDateTime = start.Plus(Duration.FromDays(30)).ToString(),
+                    MeteringGridAreaDomainId = "804",
+                    EnergySupplierMarketParticipantId = "5790001330552",
+                    SettlementVersion = "D01",
+                    Resolution = "D01",
```
RasmusGraabaek, line 118:
```suggestion
                    Resolution = "PT15M",
```

source/SubsystemTests/Drivers/B2C/Client/swaggerV3.json
```diff
@@ -145,19 +146,7 @@
...
-              "MeteredDataAdministrator",
-              "MeteredDataResponsible",
-              "BalanceResponsibleParty",
-              "ImbalanceSettlementResponsible",
-              "SystemOperator",
-              "DanishEnergyAgency",
-              "Delegated",
-              "DataHubAdministrator"
-            ],
+            "nullable": true,
```
RasmusGraabaek, line 149: Why are these nullable?

source/SubsystemTests/Drivers/B2C/Client/swaggerV3.json
```diff
@@ -168,19 +157,7 @@
...
```
RasmusGraabaek, line 160: receiver role nullable?

source/SubsystemTests/Drivers/B2C/B2CEdiDriver.cs
```diff
@@ -94,6 +97,44 @@ await webApiClient.RequestWholesaleSettlementAsync(
             .ConfigureAwait(false);
     }
 
+    public async Task RequestWholesaleSettlementTempAsync(CancellationToken cancellationToken)
```
RasmusGraabaek, line 100: Is this a valid request or just a request?

source/SubsystemTests/Drivers/B2C/B2CEdiDriver.cs
```diff
@@ -94,6 +97,44 @@ await webApiClient.RequestWholesaleSettlementAsync(
...
+    {
+        var webApiClient = await CreateWebApiClientAsync();
+
+        var start = Instant.FromUtc(2024, 09, 01, 00, 00);
```
RasmusGraabaek, line 104:
```suggestion
        var start = Instant.FromUtc(2024, 08, 31, 22, 00);
```

source/B2CWebApi/Controllers/TempRequestWholesaleSettlementController.cs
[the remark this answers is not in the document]
lasrinnil: Electricty, approved. But wont refactor it downstream. :)

RasmusGraabaek (approved):

PR #1567, opened 2025-04-02
feat: Assign forward metered data reject to the actor starting the request

RasmusGraabaek:
## Checklist
- [ ] Reference to the task


source/OutgoingMessages.IntegrationTests/OutgoingMessages/Bundling/WhenEnqueuingMeasureDataWithBundlingTests.cs
RasmusGraabaek, line 1: Rename file? If so, to what?

source/Tests/B2BApi/Functions/EnqueueMessages/Mappers/RejectedForwardMeteredDataMessageDtoMapperTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        var serviceBusMessageId = Guid.NewGuid();
+        var actor = new EnqueueActorMessagesActorV1
+        {
+            ActorNumber = "1231231231234", ActorRole = ActorRoleV1.Delegated,
+        };
+
+        var rejectData = new ForwardMeteredDataRejectedV1(
+            OriginalActorMessageId: MessageId.New().Value,
+            OriginalTransactionId: TransactionId.New().Value!,
+            ForwardedByActorNumber: ActorNumber.Create("DeleteThisYes"),
```
RasmusGraabaek, line 42: Will be removed in a later PR

source/IntegrationTests/Behaviours/IncomingRequests/GivenForwardMeteredDataV2Tests.cs
```diff
@@ -131,7 +131,7 @@ await GivenReceivedMeteredDataForMeteringPoint(
 
         // Arrange
         var requestMeteredDataForMeteringPointInputV1 = message.ParseInput<ForwardMeteredDataInputV1>();
-        var requestMeteredDataForMeteringPointAcceptedServiceBusMessage = MeteredDataForMeteringPointEventBuilder
+        var requestMeteredDataForMeteringPointAcceptedServiceBusMessage = ForwardMeteredDataResponseBuilder
```
ebbeknudsen, line 134:
```suggestion
        var forwardMeteredDataAcceptedServiceBusMessage = ForwardMeteredDataResponseBuilder
```

Same goes for other places where the rename has happened

source/B2BApi/Functions/EnqueueMessages/BRS_021/Mappers/RejectedForwardMeteredDataMessageDtoMapper.cs
```diff
@@ -0,0 +1,47 @@
...
+                OriginalTransactionIdReference: TransactionId.From(rejectedData.OriginalTransactionId),
+                TransactionId: TransactionId.New(),
+                RejectReasons: rejectedData.ValidationErrors.Select(
+                        validationError =>
+                            new RejectReason(
+                                ErrorCode: validationError.ErrorCode,
+                                ErrorMessage: validationError.Message))
+                    .ToList()));
+    }
+}
```
ebbeknudsen, line 47: I would prefer not to introduce mapper and mapper tests, unless absolutely necesarry

- What does introducing the explicit mapper solve?
- Does the mapper test guard against anything useful, or is it just extra code that needs to maintained whenever the implementation changes?
- Shouldn't the mapping be tested by our "black box" tests, like the behaviours, which we hopefully don't have to change as often, since they test the final result.

There is tons of mapping/converting/"new"'ing objects in codebases, do we really need tests for each of them? Do we want to test that `BusinessReason.FromName()` functions as expected? That `new ExternalId(..)` does what's expected? These are all "mappings" as well. Do we need a `BusinessReasonMapper` class that converts a string into a name?

source/OutgoingMessages.UnitTests/Application/OutgoingMessageFactoryTests.cs
```diff
@@ -0,0 +1,55 @@
...
+using Energinet.DataHub.EDI.OutgoingMessages.Application;
+using Energinet.DataHub.EDI.OutgoingMessages.UnitTests.Factories;
+using NodaTime;
+
+namespace Energinet.DataHub.EDI.OutgoingMessages.UnitTests.Application;
+
+public class OutgoingMessageFactoryTests
+{
+    [Fact]
+    public void Given_RejectedForwardMeteredData_When_MappingToOutgoingMessage_Then_OutgoingMessageHasExpectedValues()
```
ebbeknudsen, line 26:
```suggestion
    public void Given_RejectedForwardMeteredData_When_CreateingOutgoingMessage_Then_OutgoingMessageHasExpectedValues()
```

source/OutgoingMessages.UnitTests/Application/OutgoingMessageFactoryTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        // Arrange
+        var currentInstant = Instant.FromUtc(2024, 04, 02, 0, 0);
+        var reject = new RejectedForwardMeteredDataMessageDtoBuilder()
+            .Build();
+
+        // Act
+        var outgoingMessage = OutgoingMessageFactory.CreateMessage(
+            reject,
+            new Serializer(),
+            currentInstant);
```
ebbeknudsen, line 37: I'm unsure if this needs a test if all other `OutgoingMessageFactory` methods aren't tested. Not saying it is the wrong choice, but are we sure this isn't covered by other tests?

ebbeknudsen (approved):

PR #1572, opened 2025-04-04
chore: Update ProcessManager to v4 & fix breaking changes

ebbeknudsen:
## Description

Update to PM v4 & fix breaking changes. The deprecated `OriginalMessageId` property will be removed in a new package, and afterwards removed from EDI.

## Checklist
- [ ] Reference to the task


source/OutgoingMessages.Infrastructure/OutgoingMessages.Infrastructure.csproj
```diff
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="Energinet.DataHub.Core.Databricks.SqlStatementExecution" Version="12.0.2"/>
-    <PackageReference Include="Energinet.DataHub.ProcessManager.Orchestrations.Abstractions" Version="1.16.0" />
```
ebbeknudsen, line 10: Update to correct version before merging

- https://github.com/Energinet-DataHub/opengeh-process-manager/pull/321

RasmusGraabaek (approved):

PR #1574, opened 2025-04-05
refac: B2C Request calculated data

lasrinnil:
## Description
- Created a simpler B2C API for requesting data - some of the client logic leaked into the earlier implementation. This will have to be moved to the BFF when using the new API.

- Added versioning to the controllers handling requests for data.

## References
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/548
## Checklist
- [x] Subsystem test executed (dev_002/dev_003) https://github.com/Energinet-DataHub/dh3-environments/actions/runs/14374499068
- [x] Reference to the task


source/B2CWebApi/Factories/RequestAggregatedMeasureDataDtoFactoryV1.cs
```diff
@@ -0,0 +1,149 @@
...
+        switch (request.SettlementMethod)
+        {
+            case SettlementMethod.Flex:
+                return "D01";
+            case SettlementMethod.NonProfiled:
+                return "E02";
+        }
+
+        return null;
+    }
```
ManBearTM, line 128: suggestion: since you are already adding some convenience with the custom mapping of `MeteringPointType`, it seems a bit silly to also require specifying the `SettlementMethod` parameter when that could just as easily be derived from `MeteringPointType` (if I understand it correctly). So I think that we should either change the `MeteringPointType` to only have the 3 values (Production, Consumption, Exchange) OR keep `MeteringPointType` and then remove `SettlementMethod` as a parameter (just derive it from `MeteringPointType).

source/B2CWebApi/Factories/RequestAggregatedMeasureDataDtoFactoryV1.cs
```diff
@@ -0,0 +1,149 @@
...
+        string senderNumber,
+        string senderRole,
+        Instant now)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+
+        var senderRoleCode = MapRoleNameToCode(senderRole);
+
+        var series = new RequestAggregatedMeasureDataSeries(
+            Id: transactionId.Value,
```
RasmusGraabaek, line 45: Why are we tanking "transactionId.Value" in as an input, when we are creating the messageId at line 62

source/B2CWebApi/Models/Resolution.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Energinet.DataHub.EDI.B2CWebApi.Models;
+
+public enum Resolution
+{
+    QuarterHourly,
+    Hourly,
+    Daily,
+    Monthly,
+}
```
ManBearTM, line 23: question: it looks to me like we only have a use for the`Monthly` resolution in the frontend (either `Monthly` or `null`). does it make sense to change this enum to reflect that or do you think we will have to support other resolutions in the future? also, I assume that b2b requests can already have different resolutions other than `Monthly` or `null`, so there is also the challenge of how to map that to the "Price Type" that we use in the frontend.

source/B2CWebApi/Factories/RequestAggregatedMeasureDataDtoFactoryV1.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        if (calculationType.SettlementVersion == SettlementVersion.ThirdCorrection)
+        {
+            return "D03";
+        }
+
+        return null;
+    }
+
+    private static string MapToBusinessReasonCode(RequestAggregatedMeasureDataMarketRequestV1 requestCalculationType)
```
RasmusGraabaek, line 88:
```suggestion
    private static string MapToBusinessReasonCode(RequestAggregatedMeasureDataMarketRequestV1 request)
```

ManBearTM (approved):
Looks good, just a few comments 🙂

source/B2CWebApi/Factories/RequestWholesaleSettlementDtoFactoryV1.cs
```diff
@@ -0,0 +1,139 @@
...
+
+namespace Energinet.DataHub.EDI.B2CWebApi.Factories;
+
+public static class RequestWholesaleSettlementDtoFactoryV1
+{
+    private const string WholesaleSettlementMessageType = "D21";
+    private const string Electricity = "23";
+
+    public static RequestWholesaleSettlementDto Create(
+        TransactionId transactionId,
```
RasmusGraabaek, line 36: Reason for not using TransactionId.new() ?

source/B2CWebApi/Factories/RequestWholesaleSettlementDtoFactoryV1.cs
```diff
@@ -0,0 +1,139 @@
...
+        return calculationType.SettlementVersion switch
+        {
+            Models.SettlementVersion.FirstCorrection => SettlementVersion.FirstCorrection.Code,
+            Models.SettlementVersion.SecondCorrection => SettlementVersion.SecondCorrection.Code,
+            Models.SettlementVersion.ThirdCorrection => SettlementVersion.ThirdCorrection.Code,
+            _ => null,
+        };
+    }
+
+    private static string MapToBusinessReasonCode(RequestWholesaleSettlementMarketRequestV1 calculationType)
```
RasmusGraabaek, line 110:
```suggestion
    private static string MapToBusinessReasonCode(RequestWholesaleSettlementMarketRequestV1 request)
```

source/B2CWebApi/Factories/RequestWholesaleSettlementDtoFactoryV1.cs
```diff
@@ -0,0 +1,139 @@
...
+        return calculationType.BusinessReason switch
+        {
+            Models.BusinessReason.PreliminaryAggregation => BusinessReason.PreliminaryAggregation.Code,
+            Models.BusinessReason.BalanceFixing => BusinessReason.BalanceFixing.Code,
+            Models.BusinessReason.WholesaleFixing => BusinessReason.WholesaleFixing.Code,
+            Models.BusinessReason.Correction => BusinessReason.Correction.Code,
+            _ => throw new ArgumentOutOfRangeException(
+                nameof(calculationType),
+                calculationType,
+                "Unknown CalculationType"),
```
RasmusGraabaek, line 121:
```suggestion
                "Unknown BusinessReason"),
```

source/B2CWebApi/Factories/RequestAggregatedMeasureDataDtoFactoryV1.cs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        return requestCalculationType.BusinessReason switch
+        {
+            Models.BusinessReason.PreliminaryAggregation => BusinessReason.PreliminaryAggregation.Code,
+            Models.BusinessReason.BalanceFixing => BusinessReason.BalanceFixing.Code,
+            Models.BusinessReason.WholesaleFixing => BusinessReason.WholesaleFixing.Code,
+            Models.BusinessReason.Correction => BusinessReason.Correction.Code,
+            _ => throw new ArgumentOutOfRangeException(nameof(requestCalculationType), requestCalculationType, "Unknown CalculationType"),
```
RasmusGraabaek, line 96:
```suggestion
            _ => throw new ArgumentOutOfRangeException(nameof(requestCalculationType), requestCalculationType, "Unknown BusinessReason"),
```

source/B2CWebApi/Models/RequestAggregatedMeasureDataMarketRequestV1.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Energinet.DataHub.EDI.B2CWebApi.Models;
+
+/// <summary>
+/// Responsible for carrying the market message data from the incoming message before any data validation.
+/// </summary>
+public record RequestAggregatedMeasureDataMarketRequestV1(
+    BusinessReason BusinessReason,
+    SettlementVersion? SettlementVersion,
+    SettlementMethod SettlementMethod,
```
RasmusGraabaek, line 23:
```suggestion
    SettlementMethod? SettlementMethod,
```
Settlement method  should be null when "MeteringPointType" is production or exchange

RasmusGraabaek (commented):
Left a few comments

source/B2CWebApi/Models/Resolution.cs
RasmusGraabaek: Wholesale request support monthly or null
Where aggregated support quarterly, hourly and daily

source/B2CWebApi/Factories/V1/RequestAggregatedMeasureDataDtoFactoryV1.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        var senderRoleCode = MapRoleNameToCode(senderRole);
+
+        var series = new RequestAggregatedMeasureDataSeries(
+            Id: TransactionId.New().Value,
+            MarketEvaluationPointType: MapEvaluationPointTypeCode(request),
+            MarketEvaluationSettlementMethod: MapSettlementMethodCode(request),
+            StartDateAndOrTimeDateTime: request.StartDate.ToString(),
+            EndDateAndOrTimeDateTime: request.EndDate.ToString(),
+            MeteringGridAreaDomainId: request.GridAreaCode,
```
RasmusGraabaek, line 49:
```suggestion
            MeteringGridAreaDomainId: request.GridArea,
```

RasmusGraabaek (approved):

PR #1577, opened 2025-04-09
Chore: Upgrade to dotnet 9

RasmusGraabaek:
## Checklist
- [ ] Reference to the task


source/SubsystemTests/Drivers/Ebix/EbixDriver.cs
```diff
@@ -54,7 +53,8 @@ public EbixDriver(Uri dataHubUrlEbixUrl, EbixCredentials ebixCredentials, ITestO
 
         _ebixServiceClient = new marketMessagingB2BServiceV01PortTypeClient(binding, endpoint);
 
-        _certificate = _ebixServiceClient.ClientCredentials.ClientCertificate.Certificate = new X509Certificate2(
+        // TODO: is this correct?
```
dstenroejl, line 56: Can we remove this TODO now?

dstenroejl (commented):
One question

source/SubsystemTests/Drivers/Ebix/EbixDriver.cs
RasmusGraabaek: Yes, we can. Sorry

dstenroejl (approved):
LGTM

lanni-energinet (approved):

PR #1584, opened 2025-04-15
fix: Reduce memory usage when a bundle is peeked

MadsDue:
## Description
We are getting OutOfMemoryException when bundles are peeked. 

Operation ID: 7ed6d6e979d90aa45dca6abd35bb64f7, where we try to bundle 300 RSM009 together and end up using 1.3gb. ([our Function App has 3.5 gb memory](https://learn.microsoft.com/en-us/azure/azure-functions/functions-premium-plan?tabs=portal#available-instance-skus))

image

Memory usages:
Before
RSM009 before

After
RSM009 after

## This PR includes the following changes:

- **Refactored `FileStorageFile.cs`**  
  Replaced the use of `StreamReader` with a `StringBuilder` for reading blob content. Previously, a new `StreamReader` was created for each message in the bundle, which could lead to increased memory usage due to cumulative overhead.

- **Modified `Bundles` structure**  
  Removed `OutgoingMessages` from `Bundles`. The class now only contains the serialized content of the outgoing messages.

- **Refactored `OutgoingMessageRepository.cs`**  
  - Now streams each `fileStorageReference` individually to load the serialized content of an outgoing message, instead of loading all messages in a single chunk.  
  - Ensures the `fileStorageFile` is properly disposed of after the serialized content has been loaded.

## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/14489507972
- [x] Reference to the task => ALert

### DIFF
--- a/source/OutgoingMessages.Domain/Models/OutgoingMessages/OutgoingMessageBundle.cs
+++ b/source/OutgoingMessages.Domain/Models/OutgoingMessages/OutgoingMessageBundle.cs
@@ -27,7 +27,7 @@ public class OutgoingMessageBundle
         ActorNumber senderId,
         ActorRole senderRole,
         MessageId messageId,
-        IReadOnlyCollection<OutgoingMessage> outgoingMessages,
+        IReadOnlyCollection<string> serializedContentOfOutgoingMessages,
         MessageId? relatedToMessageId = null)
     {
         DocumentType = documentType;
@@ -37,7 +37,7 @@ public class OutgoingMessageBundle
         SenderId = senderId;
         SenderRole = senderRole;
         MessageId = messageId;
-        OutgoingMessages = outgoingMessages;
+        SerializedContentOfOutgoingMessages = serializedContentOfOutgoingMessages;
         RelatedToMessageId = relatedToMessageId;
     }
 
@@ -63,5 +63,5 @@ public class OutgoingMessageBundle
 
     public MessageId? RelatedToMessageId { get; private set; }
 
-    public IReadOnlyCollection<OutgoingMessage> OutgoingMessages { get; set; }
+    public IReadOnlyCollection<string> SerializedContentOfOutgoingMessages { get; set; }
 }
--- a/source/OutgoingMessages.Domain/Models/OutgoingMessages/OutgoingMessageBundle.cs
+++ b/source/OutgoingMessages.Domain/Models/OutgoingMessages/OutgoingMessageBundle.cs
@@ -67,5 +67,5 @@ public class OutgoingMessageBundle
 
     public IReadOnlyCollection<string> SerializedContentOfOutgoingMessages { get; set; }
 
-    public IReadOnlyList<EventId> EventIds { get; set; }
+    public IReadOnlyCollection<EventId> EventIds { get; set; }
 }
--- a/source/OutgoingMessages.Domain/Models/OutgoingMessages/OutgoingMessageBundle.cs
+++ b/source/OutgoingMessages.Domain/Models/OutgoingMessages/OutgoingMessageBundle.cs
@@ -28,6 +28,7 @@ public class OutgoingMessageBundle
         ActorRole senderRole,
         MessageId messageId,
         IReadOnlyCollection<string> serializedContentOfOutgoingMessages,
+        IReadOnlyList<EventId> eventIds,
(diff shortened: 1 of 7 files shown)


source/OutgoingMessages.Domain/Models/OutgoingMessages/OutgoingMessageBundle.cs
```diff
@@ -63,5 +65,7 @@ public OutgoingMessageBundle(
 
     public MessageId? RelatedToMessageId { get; private set; }
 
-    public IReadOnlyCollection<OutgoingMessage> OutgoingMessages { get; set; }
+    public IReadOnlyCollection<string> SerializedContentOfOutgoingMessages { get; set; }
+
+    public IReadOnlyList<EventId> EventIds { get; set; }
```
ebbeknudsen, line 70: Is there a reason for one of the properties being an `IReadOnlyCollection<>` and the other one a `IReadOnlyList<>`? As far as i remember we usually use `IReadOnlyCollection<>`.

source/OutgoingMessages.Infrastructure/Repositories/OutgoingMessages/OutgoingMessageRepository.cs
```diff
@@ -57,18 +57,26 @@ public async Task<OutgoingMessageBundle> GetAsync(PeekResult peekResult, Cancell
...
+        // To avoid a huge memory spike and allowing DownloadAndSetMessageRecordAsync to dispose of the file storage stream per message to limit memory usage.
+        await foreach (var result in _context.OutgoingMessages
+                           .AsNoTracking()
+                           .Where(x => x.AssignedBundleId == peekResult.BundleId)
+                           .Select(x => new { x.FileStorageReference, x.EventId })
+                           .AsAsyncEnumerable()
+                           .WithCancellation(cancellationToken))
+        {
+            var serializedContent = await DownloadAndSetMessageRecordAsync(result.FileStorageReference, cancellationToken).ConfigureAwait(false);
+            serializedContentOfOutgoingMessages.Add(serializedContent);
```
ebbeknudsen, line 72: Did you test how much slower getting the outgoing message content from the blob storage is now, versus how much memory this optimization saved? Since we now download one at a time, and wait for each download to complete before starting the next one. If each download takes 200ms, and the bundle has 2.000 messages, that would take `0,2s * 2000 bundles = 400 seconds = 6,67 minutes` just to download the message content.

source/OutgoingMessages.Infrastructure/Repositories/OutgoingMessages/OutgoingMessageRepository.cs
```diff
@@ -163,15 +172,13 @@ public Task<int> CountMessagesReadyToBeBundledAsync(CancellationToken cancellati
             .CountAsync(cancellationToken);
     }
 
-    private async Task DownloadAndSetMessageRecordAsync(OutgoingMessage outgoingMessage, CancellationToken cancellationToken)
+    private async Task<string> DownloadAndSetMessageRecordAsync(FileStorageReference fileStorageReference, CancellationToken cancellationToken)
```
ebbeknudsen, line 175:
```suggestion
    private async Task<string> DownloadMessageContentAsync(FileStorageReference fileStorageReference, CancellationToken cancellationToken)
```

source/OutgoingMessages.Infrastructure/Repositories/OutgoingMessages/OutgoingMessageRepository.cs
MadsDue: This is the peek performance that you already logged in the test:
Before:
image
After:
image

ebbeknudsen: But how much slower is the RSM-012 peek when the message is around 50MB?

MadsDue: Arh..
Before:
image

After:
image

ebbeknudsen (dismissed):

ebbeknudsen (approved):

PR #1589, opened 2025-04-24
feat: enable delegating outgoing rsm012

MadsDue:
## Description
Ensure we check delegation relationships for outgoing RSM012
## Checklist
- [x] Subsystem test executed (dev_002/dev_003) => https://github.com/Energinet-DataHub/dh3-environments/actions/runs/14640397250
- [ ] Reference to the task


source/OutgoingMessages.Application/UseCases/DelegateMessage.cs
```diff
@@ -47,8 +47,8 @@ public async Task<OutgoingMessage> DelegateAsync(
         // because the receiver must be the same as the one who made the request
         if (messageToEnqueue.MessageCreatedFromProcess == ProcessType.RequestWholesaleResults
             || messageToEnqueue.MessageCreatedFromProcess == ProcessType.RequestEnergyResults
-            // TODO (MWO): We skip delegation for now
-            || messageToEnqueue.MessageCreatedFromProcess == ProcessType.OutgoingMeteredDataForMeteringPoint)
+            // Acknowledgement messages are not delegated
```
RasmusGraabaek, line 50: Should we change "messageToEnqueue.MessageCreatedFromProcess", such that we only consider document types and not the process?

MadsDue: Maybe, but I still think it will be moved from EDI into PM in the long term.

PR #1599, opened 2025-05-01
fix: set max request body size mb in b2b app

MadsDue:
## Description
fix: Request body too large. The max request body size is 30000000 bytes.

</body>
</html>
## Checklist
- [ ] Reference to the task

### DIFF
--- a/source/B2BApi/HostFactory.cs
+++ b/source/B2BApi/HostFactory.cs
@@ -31,7 +31,9 @@ using Energinet.DataHub.EDI.IntegrationEvents.Infrastructure.Extensions.Dependen
 using Energinet.DataHub.EDI.MasterData.Infrastructure.Extensions.DependencyInjection;
 using Energinet.DataHub.EDI.Outbox.Infrastructure;
 using Energinet.DataHub.EDI.OutgoingMessages.Infrastructure.Extensions.DependencyInjection;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.IdentityModel.Tokens;
 using OutboxContext = Energinet.DataHub.EDI.Outbox.Infrastructure.OutboxContext;
@@ -126,7 +128,13 @@ public static class HostFactory
                         .AddOutboxRetention()
 
                         // Enqueue messages from PM (using Edi Topic)
-                        .AddEnqueueActorMessagesFromProcessManager(defaultAzureCredential);
+                        .AddEnqueueActorMessagesFromProcessManager(defaultAzureCredential)
+
+                        // Configure Kestrel options
+                        .Configure<KestrelServerOptions>(options =>
+                        {
+                            options.Limits.MaxRequestBodySize = 50 * 1024 * 1024; // 50mb
+                        });
                 })
             .ConfigureLogging(
                 (hostingContext, logging) =>


source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
```diff
@@ -93,69 +78,136 @@ protected override async Task<IncomingMarketMessageParserResult> ParseMessageAsy
...
+        foreach (var transactionElement in transactionElements.EnumerateArray())
+        {
+            // This validates a full document that contains only a single transaction.
+            // By isolating one transaction per validation, we reduce peak memory usage
+            // and allow memory to be reused between validations.
+            using var documentWithOneTransaction = GetDocumentWithOneSeriesElement(transactionElement, headerElement);
+            if (!IsValid(documentWithOneTransaction, schemaResult))
+                return null;
+
+            var header = ParseHeader(documentWithOneTransaction!);
```
ebbeknudsen, line 106: Wouldn't the header be the same for all transactions? So this could be outside the loop.

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
```diff
@@ -35,44 +35,29 @@ public abstract class JsonMessageParserBase(JsonSchemaProvider schemaProvider) :
...
+        if (document == null)
+            return new IncomingMarketMessageParserResult(_validationErrors!.ToArray());
 
-        if (IsValid(document, schemaResult) == false)
-        {
-            return new IncomingMarketMessageParserResult(ValidationErrors.ToArray());
-        }
+        var transactions = ValidateAndParseTransactions(document, schemaResult);
+        if (transactions == null)
+            return new IncomingMarketMessageParserResult(_validationErrors!.ToArray());
```
ebbeknudsen, line 58: Would it make sense to parse the header before parsing the transactions? Since if the header is invalid, then we don't have to use memory to parse the potential "big" transactions.

Then we could also pass the header information to the `ValidateAndParseTransactions()` method, so we don't have to call `GetHeaderElement()` or `ParseHeader()` in that method, and thus not having to read/parse the header multiple times.

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
```diff
@@ -93,69 +78,136 @@ protected override async Task<IncomingMarketMessageParserResult> ParseMessageAsy
...
+            // This validates a full document that contains only a single transaction.
+            // By isolating one transaction per validation, we reduce peak memory usage
+            // and allow memory to be reused between validations.
+            using var documentWithOneTransaction = GetDocumentWithOneSeriesElement(transactionElement, headerElement);
+            if (!IsValid(documentWithOneTransaction, schemaResult))
+                return null;
+
+            var header = ParseHeader(documentWithOneTransaction!);
+            transactions.Add(ParseTransaction(
+                documentWithOneTransaction!.RootElement.GetProperty(HeaderElementName).GetProperty(SeriesElementName).EnumerateArray().First(),
```
ebbeknudsen, line 108: We should be able to use `.Single()` here, since we know there is only one transaction in the temporary document.

```suggestion
                documentWithOneTransaction!.RootElement.GetProperty(HeaderElementName).GetProperty(SeriesElementName).EnumerateArray().Single(),
```

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
```diff
@@ -93,69 +78,136 @@ protected override async Task<IncomingMarketMessageParserResult> ParseMessageAsy
...
+    }
+
+    private JsonDocument? GetDocumentWithOneSeriesElement(
+        JsonElement seriesElement,
+        JsonNode headerElement)
+    {
+        JsonDocument? documentWithOneTransaction = null;
+        try
+        {
+            var transactionNode = JsonNode.Parse(seriesElement.GetRawText());
```
ebbeknudsen, line 146: Is it really necesarry to read the entire element's text just to parse it into another `Json` object? Can we just cast it or similar and thus reuse the `seriesElement`, without having to read the entire content into memory as a string aswell.

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
```diff
@@ -93,69 +78,136 @@ protected override async Task<IncomingMarketMessageParserResult> ParseMessageAsy
...
+            headerElement[SeriesElementName] = new JsonArray(transactionNode);
+            var modifiedJson = headerElement.Parent!.ToJsonString();
+            documentWithOneTransaction = JsonDocument.Parse(modifiedJson);
```
ebbeknudsen, line 149: Again, is it not possible to convert this without having to write the entire element to a string just to parse it back to a json object?

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
```diff
@@ -35,44 +35,29 @@ public abstract class JsonMessageParserBase(JsonSchemaProvider schemaProvider) :
     private const string SenderRoleElementName = "sender_MarketParticipant.marketRole.type";
     private const string ReceiverRoleElementName = "receiver_MarketParticipant.marketRole.type";
     private const string CreatedDateElementName = "createdDateTime";
+    private const string BusinessSectorTypeElementName = "businessSector.type";
     private readonly JsonSchemaProvider _schemaProvider = schemaProvider;
 
+    private List<ValidationError>? _validationErrors;
```
dstenroejl, line 41: There is something odd with the lifecycle of this list, and the lifecycle/usage of the class. E.g. the list is created the first time we encounter an error; but what if I called some of the methods multiple times; and can they be called in parallel(?)
Should the list be cleared if I call methods multiple times (so I don't add "old" errors with "new"), or should a state pattern ensure I cannot call them multiple times, or in parallel.
Basically what I'm saying is the lifecycle of this list, should follow the lifecycle of the "ParseMessageAsync" method, because it is in the execution path for this method, that the list is used.

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
```diff
@@ -93,69 +78,136 @@ protected override async Task<IncomingMarketMessageParserResult> ParseMessageAsy
...
-    private bool IsValid(JsonDocument jsonDocument, JsonSchema schema)
+    private bool IsValid(JsonDocument? jsonDocument, JsonSchema schema)
     {
-        var result = schema.Evaluate(jsonDocument, new EvaluationOptions() { OutputFormat = OutputFormat.Hierarchical, });
-        if (result.IsValid == false)
+        if (jsonDocument is null || !jsonDocument.RootElement.EnumerateObject().Any())
         {
-            FindErrorsForInvalidEvaluation(result);
+            AddValidationError("Document is empty");
+            return false;
```
ebbeknudsen, line 179: Wouldn't this be caught in the `schema.Evaluate` below? Thus no need to read the entire json document here (if it is a large file).

Alternatively we could move it below the `schema.Evaluate` and thus we only need to read the entire document if there is no validation errors in `schema.Evaluate()`

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
```diff
@@ -93,69 +78,136 @@ protected override async Task<IncomingMarketMessageParserResult> ParseMessageAsy
...
         }
 
-        if (!jsonDocument.RootElement.EnumerateObject().Any())
+        var result = schema.Evaluate(jsonDocument, new EvaluationOptions { OutputFormat = OutputFormat.Hierarchical });
```
ebbeknudsen, line 182: `schema.Evaluate()` also supports `JsonElement` or `JsonNode`, so maybe we don't have to change the incomming transaction to a `JsonDocument jsonDocument` in the previous code.

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
```diff
@@ -35,44 +35,29 @@ public abstract class JsonMessageParserBase(JsonSchemaProvider schemaProvider) :
...
-        {
-            document = await JsonDocument.ParseAsync(marketMessage.Stream, cancellationToken: cancellationToken).ConfigureAwait(false);
-        }
-        catch (JsonException exception)
-        {
-            AddValidationError(exception.Message);
-            return new IncomingMarketMessageParserResult(ValidationErrors.ToArray());
-        }
+        var document = await TryParseJsonDocumentAsync(marketMessage, cancellationToken).ConfigureAwait(false);
+        if (document == null)
```
ebbeknudsen, line 53: Could we dispose the `marketMessage` stream after it is read here?

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
```diff
@@ -93,69 +78,136 @@ protected override async Task<IncomingMarketMessageParserResult> ParseMessageAsy
...
 
-    private static string GetJsonDateStringWithoutQuotes(JsonElement element)
+    private static string? GetBusinessType(JsonElement element)
     {
-        return element.ToString().Trim('"');
+        return element.TryGetProperty("businessSector.type", out var property) ? property.GetProperty("value").GetString() : null;
     }
 
-    private static string? GetBusinessType(JsonElement element)
+    private List<IIncomingMessageSeries>? ValidateAndParseTransactions(
```
dstenroejl, line 86:
```suggestion
    private IReadOnlyCollection<IIncomingMessageSeries>? ValidateAndParseTransactions(
```

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
MadsDue: Good catch. 

We still need to ensure the document is valid before parsing it. But we only need the senderId in this case, so I’ve refactored the code accordingly and only doing it once, because the header is the same as you mentioned.

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
```diff
@@ -35,44 +35,29 @@ public abstract class JsonMessageParserBase(JsonSchemaProvider schemaProvider) :
...
-            document = await JsonDocument.ParseAsync(marketMessage.Stream, cancellationToken: cancellationToken).ConfigureAwait(false);
-        }
-        catch (JsonException exception)
-        {
-            AddValidationError(exception.Message);
-            return new IncomingMarketMessageParserResult(ValidationErrors.ToArray());
-        }
+        var document = await TryParseJsonDocumentAsync(marketMessage, cancellationToken).ConfigureAwait(false);
+        if (document == null)
+            return new IncomingMarketMessageParserResult(_validationErrors!.ToArray());
```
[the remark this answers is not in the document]
MadsDue: I have refactored it as you suggested :) 

TryParseJsonDocumentAsync added a validation error if it can't parse the document :)

source/IncomingMessages.IntegrationTests/IncomingMessages/GivenIncomingMeteredDataForMeteringMessageTests.cs
```diff
@@ -663,6 +669,70 @@ public async Task When_BusinessTypeIsNotAllowed_Then_ExpectedValidationError()
         messageParser.ParserResult.Success.Should().BeFalse();
     }
 
+    [Theory]
+    [MemberData(nameof(GetAllDocumentFormat))]
+    public async Task AndGiven_EachTransactionHasOneYearDataWithQuarterlyResolution_When_MessageSizeIs50mb_Then_ValidationSucceedAndWithinAllowedExecutionTimeAndMemoryUsage(DocumentFormat documentFormat)
```
dstenroejl, line 674: Nice with a test. :)

dstenroejl (commented):
Observations mentioned

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
MadsDue: The Evaludation method acceptes an JsonNode, so it looks like we dont have to parse it back again :D

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
MadsDue: I have renamed the ValidateAndParseTransactions method and move the ParseHeader() into it.

source/IncomingMessages.Domain/MessageParsers/JsonMessageParserBase.cs
MadsDue: Good catch, I agree with you on this and have refactored it :)

ebbeknudsen (approved):
LGTM

dstenroejl (approved):
LGTM

PR #1600, opened 2025-05-01
feat: Authentication for subsystem http requests

RasmusGraabaek:
## Checklist
- [ ] Reference to the task

### DIFF
--- a/source/B2BApi.AppTests/SubsystemHttpTrigger/EnqueueHttpEndpointTests.cs
+++ b/source/B2BApi.AppTests/SubsystemHttpTrigger/EnqueueHttpEndpointTests.cs
@@ -67,7 +67,7 @@ public class EnqueueHttpEndpointTests : IAsyncLifetime
     }
 
     [Fact]
-    public async Task Given_SubsystemRequestWithInValidToken_When_Requesting_Then_RejectedRequest()
+    public async Task Given_SubsystemRequestWithInvalidToken_When_Requesting_Then_RejectedRequest()
     {
         // Arrange
         var token = "InvalidToken";
@@ -81,7 +81,7 @@ public class EnqueueHttpEndpointTests : IAsyncLifetime
 
     private HttpRequestMessage CreateHttpRequest(string token)
     {
-        var request = new HttpRequestMessage(HttpMethod.Get, "monitor/live");
+        var request = new HttpRequestMessage(HttpMethod.Get, "api/monitor/live");
 
         request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", token);
         return request;
--- a/source/B2BApi.AppTests/SubsystemHttpTrigger/EnqueueHttpEndpointTests.cs
+++ b/source/B2BApi.AppTests/SubsystemHttpTrigger/EnqueueHttpEndpointTests.cs
@@ -1,0 +1,89 @@
+﻿// Copyright 2020 Energinet DataHub A/S
+//
+// Licensed under the Apache License, Version 2.0 (the "License2");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Net;
+using System.Net.Http.Headers;
+using Azure.Core;
+using Azure.Identity;
+using Energinet.DataHub.EDI.B2BApi.AppTests.Fixtures;
+using FluentAssertions;
+using Xunit;
+using Xunit.Abstractions;
+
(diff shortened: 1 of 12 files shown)


source/B2BApi.AppTests/SubsystemHttpTrigger/EnqueueHttpEndpointTests.cs
```diff
@@ -0,0 +1,107 @@
...
+using Energinet.DataHub.EDI.B2BApi.Authentication;
+using Energinet.DataHub.EDI.BuildingBlocks.Domain.Models;
+using Energinet.DataHub.EDI.IntegrationTests.Infrastructure.Authentication.MarketActors;
+using FluentAssertions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Energinet.DataHub.EDI.B2BApi.AppTests.SubsystemHttpTrigger;
+
+[Collection(nameof(B2BApiAppCollectionFixture))]
```
RasmusGraabaek, line 27: Update this, when the enqueue trigger is merged

source/B2BApi/Configuration/Middleware/FunctionContextExtensions.cs
```diff
@@ -29,7 +31,9 @@ internal static bool IsProtectedHttpTrigger(this FunctionContext context)
         var isHealthCheckEndpoint = context.FunctionDefinition.Name == "HealthCheck";
         var isDurableFunctionMonitorEndpoint = context.FunctionDefinition.PathToAssembly.EndsWith("durablefunctionsmonitor.dotnetisolated.core.dll");
 
-        return isHttpTrigger && !isHealthCheckEndpoint && !isDurableFunctionMonitorEndpoint;
+        var hasAttribute = HasAuthorizeAttribute(context.FunctionDefinition.EntryPoint);
+
+        return isHttpTrigger && !isHealthCheckEndpoint && !isDurableFunctionMonitorEndpoint && !hasAttribute;
```
dstenroejl, line 36: I had some trouble understanding this logic, until I realized that its the name of the method that is "inprecise". Maybe it should be called "IsMarketActorProtectedEndpoint". Since an endpoint that we add "Authorize" on is also "a protected endpoint".

source/B2BApi/Extensions/DependencyInjection/SubsystemAuthenticationExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+                options.Authority = authenticationOptions.Issuer;
+            });
+
+        return services;
+    }
+
+    private static void GuardAuthenticationOptions(AuthenticationOptions authenticationOptions)
+    {
+        if (string.IsNullOrWhiteSpace(authenticationOptions.ApplicationIdUri))
+            throw new InvalidConfigurationException($"Missing '{nameof(AuthenticationOptions.ApplicationIdUri)}'.");
```
dstenroejl, line 58: I have used this exception in some places as well, but it is actually a Microsoft exception for specific scenarious, so I have started dropping its use and instead just use `InvalidOperationException` (like on line 40).

source/B2BApi/Extensions/DependencyInjection/SubsystemAuthenticationExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+
+public static class SubsystemAuthenticationExtensions
+{
+    /// <summary>
+    /// Register services necessary for enabling an Azure Function App (isolated worker model)
+    /// to use JWT Bearer authentication for HttpTrigger's. This is using DarkLoop Authorization extension.
+    ///
+    /// Expects <see cref="AuthenticationOptions"/> has been configured in <see cref="AuthenticationOptions.SectionName"/>.
+    /// </summary>
+    internal static IServiceCollection AddSubsystemAuthentication(this IServiceCollection services, IConfiguration configuration)
```
dstenroejl, line 31: Maybe just describe what we mean by "subsystem authentication". E.g. we could briefly mention "this is one subsystem calling another subsystem with a token"

source/B2BApi/Extensions/Options/AuthenticationOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.ComponentModel.DataAnnotations;
+
+namespace Energinet.DataHub.EDI.B2BApi.Extensions.Options;
+
+/// <summary>
+/// Contains options for validating the JWT bearer tokens that must be sent as
+/// part of any http request for protected http endpoints.
```
dstenroejl, line 21: Since we have multiple http request scenarious in EDI (an B2B actor calling an http endpoint or another subsystem calling an http endpoint), then maybe we should mention here that this is for "subsystem to subsystem communication" and not any http request. We could even consider if the section name should be more specific and be something like "SubsystemAuth" or similar.

dstenroejl (commented):
Nice. Added some suggestions

source/B2BApi/Configuration/Middleware/FunctionContextExtensions.cs
RasmusGraabaek: Renamed to "IsActorProtectedEndpoint"

dstenroejl (approved):
LGTM

PR #1603, opened 2025-05-08
test: Mock dependencies when running subsystem tests

ebbeknudsen:
## Description

This PR updates test dependencies to use special test UUIDs and improves testability by mocking dependencies in subsystem tests. Key changes include:

- Replacing Guid.ToString() calls with Guid.ToTestMessageUuid() for consistent test message identifiers.
- Changing return types from Guid to string for test methods where message identifiers are returned.
- Renaming and updating configuration options from IncomingMessagesQueueOptions to IncomingMessagesOptions and adding a new ProcessManagerClientFactory to support mock dependencies.

## References

- Infrastructure PR: https://github.com/Energinet-DataHub/dh3-infrastructure/pull/3564
- D002 run: https://github.com/Energinet-DataHub/dh3-environments/actions/runs/14910113600/job/41882830343
- https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/740

## Checklist
- [x] Reference to the task

### DIFF
--- a/source/IncomingMessages.Infrastructure/Configuration/Options/IncomingMessagesOptions.cs
+++ b/source/IncomingMessages.Infrastructure/Configuration/Options/IncomingMessagesOptions.cs
@@ -16,10 +16,17 @@ using System.ComponentModel.DataAnnotations;
 
 namespace Energinet.DataHub.EDI.IncomingMessages.Infrastructure.Configuration.Options;
 
-public class IncomingMessagesQueueOptions
+public class IncomingMessagesOptions
 {
     public const string SectionName = "IncomingMessages";
 
     [Required]
     public string QueueName { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Allow using mock dependencies for tests. This is used by our subsystem tests, to avoid polluting other subsystems,
+    /// and thus must only be enabled in environments where subsystem tests are executed.
+    /// </summary>
+    [Required]
+    public bool AllowMockDependenciesForTests { get; set; } = false;
 }
--- a/source/IncomingMessages.Infrastructure/ProcessManager/IProcessManagerClientFactory.cs
+++ b/source/IncomingMessages.Infrastructure/ProcessManager/IProcessManagerClientFactory.cs
@@ -1,0 +1,33 @@
+﻿// Copyright 2020 Energinet DataHub A/S
+//
+// Licensed under the Apache License, Version 2.0 (the "License2");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Energinet.DataHub.EDI.BuildingBlocks.Domain.Extensions;
+using Energinet.DataHub.EDI.IncomingMessages.Infrastructure.Configuration.Options;
+using Energinet.DataHub.ProcessManager.Client;
+
+namespace Energinet.DataHub.EDI.IncomingMessages.Infrastructure.ProcessManager;
+
+public interface IProcessManagerClientFactory
+{
+    /// <summary>
+    /// Create a process manager message client.
+    /// <remarks>
+    /// If <see cref="IncomingMessagesOptions.AllowMockDependenciesForTests"/> is true, and the
+    /// given <paramref name="actorMessageId"/> is a test id, a mock client will be returned. See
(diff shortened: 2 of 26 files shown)


source/IncomingMessages.Infrastructure/ProcessManager/IProcessManagerClientFactory.cs
```diff
@@ -0,0 +1,33 @@
...
+using Energinet.DataHub.EDI.IncomingMessages.Infrastructure.Configuration.Options;
+using Energinet.DataHub.ProcessManager.Client;
+
+namespace Energinet.DataHub.EDI.IncomingMessages.Infrastructure.ProcessManager;
+
+public interface IProcessManagerClientFactory
+{
+    /// <summary>
+    /// Create a process manager message client.
+    /// <remarks>
```
dstenroejl, line 25: I believe the "remarks" section must be outside the "summary"; otherwise it will just be shown as one text section.

source/IncomingMessages.Infrastructure/ProcessManager/ProcessManagerClientFactory.cs
```diff
@@ -0,0 +1,75 @@
...
+using Energinet.DataHub.ProcessManager.Client;
+using Microsoft.Extensions.Options;
+
+namespace Energinet.DataHub.EDI.IncomingMessages.Infrastructure.ProcessManager;
+
+/// <summary>
+/// Create process manager clients. Can return a mock client if
+/// <see cref="IncomingMessagesOptions.AllowMockDependenciesForTests"/> is true.
+/// </summary>
+public class ProcessManagerClientFactory(
```
dstenroejl, line 27:
```suggestion
internal class ProcessManagerClientFactory(
```

source/IncomingMessages.Infrastructure/Configuration/Options/IncomingMessagesOptions.cs
```diff
@@ -16,10 +16,17 @@
 
 namespace Energinet.DataHub.EDI.IncomingMessages.Infrastructure.Configuration.Options;
 
-public class IncomingMessagesQueueOptions
+public class IncomingMessagesOptions
```
dstenroejl, line 19: The usual pattern for placing options is "Extensions/Options", see https://energinet.atlassian.net/wiki/spaces/D3/pages/767459361/Application+Development+Guidelines#Options-Conventions

source/IncomingMessages.Infrastructure/ProcessManager/IProcessManagerClientFactory.cs
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Energinet.DataHub.EDI.BuildingBlocks.Domain.Extensions;
+using Energinet.DataHub.EDI.IncomingMessages.Infrastructure.Configuration.Options;
+using Energinet.DataHub.ProcessManager.Client;
+
+namespace Energinet.DataHub.EDI.IncomingMessages.Infrastructure.ProcessManager;
+
+public interface IProcessManagerClientFactory
```
dstenroejl, line 21: Not sure if the class name is intentional, or it should contain "Message" in the name. Like `IProcessManagerMessageClientFactory`.

dstenroejl (commented):
A few input/observations.

source/IncomingMessages.Infrastructure/ProcessManager/IProcessManagerClientFactory.cs
ebbeknudsen: As far as i know it should be inside the summary. That is atleast what we've done in a lot of other places:
image


Or maybe it just doesn't matter 🤔 . I can see they place it outside the tag in their documentation: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#d312-remarks

ebbeknudsen: I've moved it outside the summary tag.

source/IncomingMessages.Infrastructure/Configuration/Options/IncomingMessagesOptions.cs
ebbeknudsen: We havn't really been following those in EDI, many of the other options are in a `Configuration` folder (like this one is, i've only renamed the class). Which also makes much more sense to me.

source/IncomingMessages.Infrastructure/ProcessManager/IProcessManagerClientFactory.cs
ebbeknudsen: My initial thought was that it could create both `IProcessManagerClient` and `IProcessManagerHttpClient`'s, but I've renamed it to `IProcessManagerHttpClientFactory` since i don't think we're going to use the http client in EDI.

PR #1606, opened 2025-05-08
test: Add subsystem test for enqueue http endpoint

RasmusGraabaek:
## References
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/741

## Checklist
- [ ] ~Should the change be behind a feature flag?~
- [ ] ~Can the feature be meaningfully disabled or circumvented if there are issues (e.g., database-breaking changes)?~
- [x] Subsystem test executed (dev_002/dev_003) https://github.com/Energinet-DataHub/dh3-environments/actions/runs/14925535318
- [x] Reference to the task

### DIFF
--- a/source/SubsystemTests/Drivers/EdiDriver.cs
+++ b/source/SubsystemTests/Drivers/EdiDriver.cs
@@ -37,10 +37,14 @@ internal sealed class EdiDriver
     private readonly AsyncLazy<HttpClient>? _subsystemHttpClient;
     private readonly ITestOutputHelper _logger;
 
-    public EdiDriver(IDurableClient durableClient, AsyncLazy<HttpClient> b2BB2BHttpClient, ITestOutputHelper logger,  AsyncLazy<HttpClient>? subsystemHttpClient = null)
+    public EdiDriver(
+        IDurableClient durableClient,
+        AsyncLazy<HttpClient> b2BHttpClient,
+        ITestOutputHelper logger,
+        AsyncLazy<HttpClient>? subsystemHttpClient = null)
     {
         _durableClient = durableClient;
-        _b2BHttpClient = b2BB2BHttpClient;
+        _b2BHttpClient = b2BHttpClient;
         _subsystemHttpClient = subsystemHttpClient;
         _logger = logger;
     }
--- a/source/SubsystemTests/Drivers/EdiDriver.cs
+++ b/source/SubsystemTests/Drivers/EdiDriver.cs
@@ -220,7 +220,7 @@ internal sealed class EdiDriver
 
         var response = await httpClient.SendAsync(request);
 
-        response.EnsureSuccessStatusCode();
+        await response.EnsureSuccessStatusCodeWithLogAsync(_logger);
     }
 
     private static async Task<(Guid MessageId, string Content)> GetRequestWholesaleSettlementContentAsync(
--- a/source/SubsystemTests/Drivers/EdiDriver.cs
+++ b/source/SubsystemTests/Drivers/EdiDriver.cs
@@ -16,11 +16,13 @@ using System.Diagnostics;
 using System.Net;
 using System.Net.Http.Headers;
 using System.Text;
+using System.Text.Json;
 using System.Xml;
 using Energinet.DataHub.Core.DurableFunctionApp.TestCommon.DurableTask;
 using Energinet.DataHub.EDI.BuildingBlocks.Domain.Models;
 using Energinet.DataHub.EDI.SubsystemTests.Exceptions;
 using Energinet.DataHub.EDI.SubsystemTests.Logging;
+using Energinet.DataHub.ProcessManager.Components.Abstractions.EnqueueActorMessages;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask;
 using Nito.AsyncEx;
 using NodaTime;
@@ -31,13 +33,15 @@ namespace Energinet.DataHub.EDI.SubsystemTests.Drivers;
 internal sealed class EdiDriver
 {
     private readonly IDurableClient _durableClient;
-    private readonly AsyncLazy<HttpClient> _httpClient;
+    private readonly AsyncLazy<HttpClient> _b2BHttpClient;
+    private readonly AsyncLazy<HttpClient>? _subsystemHttpClient;
     private readonly ITestOutputHelper _logger;
 
-    public EdiDriver(IDurableClient durableClient, AsyncLazy<HttpClient> b2bHttpClient, ITestOutputHelper logger)
+    public EdiDriver(IDurableClient durableClient, AsyncLazy<HttpClient> b2BB2BHttpClient, ITestOutputHelper logger,  AsyncLazy<HttpClient>? subsystemHttpClient = null)
     {
         _durableClient = durableClient;
-        _httpClient = b2bHttpClient;
+        _b2BHttpClient = b2BB2BHttpClient;
+        _subsystemHttpClient = subsystemHttpClient;
         _logger = logger;
     }
 
@@ -107,7 +111,7 @@ internal sealed class EdiDriver
 
     internal async Task<string> RequestAggregatedMeasureDataXmlAsync(XmlDocument payload, string? token = null)
     {
-        var b2bClient = await _httpClient;
+        var b2bClient = await _b2BHttpClient;
         using var request = new HttpRequestMessage(HttpMethod.Post, "v1.0/cim/requestaggregatedmeasuredata");
         if (token is not null) request.Headers.Authorization = new AuthenticationHeaderValue("bearer", token);
         request.Content = new StringContent(payload.OuterXml, Encoding.UTF8, "application/xml");
@@ -120,7 +124,7 @@ internal sealed class EdiDriver
 
     internal async Task<Guid> RequestWholesaleSettlementAsync(bool withSyncError, CancellationToken cancellationToken)
     {
-        var b2bClient = await _httpClient;
+        var b2bClient = await _b2BHttpClient;
         using var request = new HttpRequestMessage(HttpMethod.Post, "v1.0/cim/requestwholesalesettlement");
         var contentType = "application/json";
         var requestContent = await GetRequestWholesaleSettlementContentAsync(withSyncError, cancellationToken)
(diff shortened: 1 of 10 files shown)


source/SubsystemTests/Drivers/EdiDriver.cs
```diff
@@ -198,6 +206,23 @@ internal async Task WaitForOrchestrationCompletedAsync(string orchestrationInsta
...
+        var json = JsonSerializer.Serialize(data, data.GetType());
+
+        using var request = new HttpRequestMessage(HttpMethod.Post, "api/enqueue/" + data.Route);
+        request.Content = new StringContent(json, Encoding.UTF8, "application/json");
+
+        var httpClient = await _subsystemHttpClient;
+
+        var response = await httpClient.SendAsync(request);
+
+        response.EnsureSuccessStatusCode();
```
ebbeknudsen, line 223: use `.EnsureSuccessStatusCodeWithLog()` to enable logging the actual response if it fails, instead of just throwing an exception.

source/SubsystemTests/Drivers/EdiDriver.cs
```diff
@@ -198,6 +206,23 @@ internal async Task WaitForOrchestrationCompletedAsync(string orchestrationInsta
...
+    }
```
ebbeknudsen, line 224: It it on purpose

That you wrote code

Where each line has extra spacing

Instead of trying to group lines that belong together

And only use spacing for logical separation? 😆

source/SubsystemTests/Dsl/EnqueueActorMessagesHttpDsl.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        var processManagerReceiver = new EnqueueCalculatedMeasurementsHttpV1.Actor(
+            receiver.ActorNumber.ToProcessManagerActorNumber(),
+            receiver.ActorRole.ToProcessManagerActorRole());
+
+        var start = Instant.FromUtc(2025, 01, 31, 23, 00, 00);
+
+        var receiversWithMeasurements = new EnqueueCalculatedMeasurementsHttpV1.ReceiversWithMeasurements(
+            Receivers: new[] { processManagerReceiver },
+            RegistrationDateTime: start.ToDateTimeUtc(),
```
ebbeknudsen, line 53:
```suggestion
            RegistrationDateTime: start.ToDateTimeOffset(),
```

source/SubsystemTests/Dsl/EnqueueActorMessagesHttpDsl.cs
```diff
@@ -0,0 +1,118 @@
...
+            StartDateTime: start.ToDateTimeUtc(),
```
ebbeknudsen, line 54:
```suggestion
            StartDateTime: start.ToDateTimeOffset()(),
```

source/SubsystemTests/Dsl/EnqueueActorMessagesHttpDsl.cs
```diff
@@ -0,0 +1,118 @@
...
+            EndDateTime: start.Plus(Duration.FromMinutes(15)).ToDateTimeUtc(),
```
ebbeknudsen, line 55:
```suggestion
            EndDateTime: start.Plus(Duration.FromMinutes(15)).ToDateTimeOffset()(),
```

source/SubsystemTests/Dsl/EnqueueActorMessagesHttpDsl.cs
```diff
@@ -0,0 +1,118 @@
...
+                    QuantityQuality: Quality.AsProvided),
+            },
+            GridAreaCode: "804");
+
+        var message = new EnqueueCalculatedMeasurementsHttpV1(
+            OrchestrationInstanceId: Guid.NewGuid(),
+            TransactionId: Guid.NewGuid(),
+            MeteringPointId: meteringPointId,
+            MeteringPointType: MeteringPointType.Consumption,
+            Resolution: Resolution.QuarterHourly,
```
ebbeknudsen, line 70: I've been informated that calculated measurements are always in `Hourly` resolution, so we should probably update our tests to use that as well, since that is more "correct".

source/SubsystemTests/Dsl/EnqueueActorMessagesHttpDsl.cs
```diff
@@ -0,0 +1,118 @@
...
+                    EnergyQuantity: 019293m,
+                    QuantityQuality: Quality.AsProvided),
+            },
+            GridAreaCode: "804");
+
+        var message = new EnqueueCalculatedMeasurementsHttpV1(
+            OrchestrationInstanceId: Guid.NewGuid(),
+            TransactionId: Guid.NewGuid(),
+            MeteringPointId: meteringPointId,
+            MeteringPointType: MeteringPointType.Consumption,
```
ebbeknudsen, line 69: Metering point type should be either `ElectricalHeating`, `NetConsumption` or `CapacitySettlement`.

source/SubsystemTests/Dsl/EnqueueActorMessagesHttpDsl.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        var numberOfRetries = 2;
+        var foundExpectedMessage = false;
+
+        for (var i = 0; i < numberOfRetries; i++)
+        {
+            var (peekResponse, dequeueResponse) = await _ediDriver.PeekMessageAsync(
+                documentFormat: DocumentFormat.Json,
+                messageCategory: MessageCategory.MeasureData,
+                timeout: timeout);
```
ebbeknudsen, line 89: `.PeekMessageAsync()` already handles peeking until a message is found. I don't think we should use a different pattern here than what we do in our other subsystem tests, where we peek until a message is found, and then check if it is the correct message.

If we want a pattern were we peek until the correct message is found, i think we should implement that pattern in the EDI driver, so other subsystem tests also benefit from it.

source/SubsystemTests/Dsl/EnqueueActorMessagesHttpDsl.cs
```diff
@@ -0,0 +1,118 @@
...
+            var contentString = await peekResponse.Content.ReadAsStringAsync();
+
+            if (IsCorrectDocumentType(contentString) && IsExpectedMeteringPointType(contentString, meteringPointId))
+            {
+                foundExpectedMessage = true;
+                break;
+            }
+        }
+
+        foundExpectedMessage.Should().BeTrue($"because we expected to peek the expected message within {numberOfRetries} retries");
```
ebbeknudsen, line 100: We must use `Assert.True()` instead of fluent assertions. 😢

source/SubsystemTests/Dsl/EnqueueActorMessagesHttpDsl.cs
```diff
@@ -0,0 +1,118 @@
...
+    private bool IsExpectedMeteringPointType(string content, string meteringPointId)
+    {
+        var expectedMeteringPointFormated = string.Empty
+                                            + "        \"marketEvaluationPoint.mRID\": {\r\n"
+                                            + "          \"codingScheme\": \"A10\",\r\n"
+                                            + $"          \"value\": \"{meteringPointId}\"\r\n"
+                                            + "        },";
+
+        return content.Contains(expectedMeteringPointFormated) == true;
+    }
```
ebbeknudsen, line 117: A pattern where we can use test assertions is much prefered, since those will write the actual content in the test result, if the content does not contain the expected values.

source/SubsystemTests/SubsystemTestFixture.cs
```diff
@@ -158,6 +159,15 @@ public SubsystemTestFixture()
             "MARKET_PARTICIPANT_URI",
             defaultValue: "https://app-webapi-markpart-u-001.azurewebsites.net"));
 
+        EdiFunctionAppUri = new Uri(
+            GetConfigurationValue<string>(
+                root,
+                "func-edi-api-base-url"));
+
+        EdiApplicationId = GetConfigurationValue<string>(
```
ebbeknudsen, line 167:
```suggestion
        EdiApplicationIdUri = GetConfigurationValue<string>(
```

We've decided to call it "application id uri" (since that is what it is called in Azure), so let is keep that naming to avoid confusion.

source/SubsystemTests/Tests/WhenEnqueueActorMessagesViaHttpTests.cs
```diff
@@ -0,0 +1,54 @@
...
+            output,
+            fixture.SubsystemHttpClient);
+
+        _enqueueActorMessagesHttpDsl = new EnqueueActorMessagesHttpDsl(ediDriver);
+    }
+
+    [Fact]
+    public async Task Actor_can_peek_messages_enqueued_via_http_endpoint()
+    {
+        var meteringPointId = "TestMeteringPointId";
```
ebbeknudsen, line 45: Is this a valid metering point id? Isn't the metering point id more like a GLN number?

source/SubsystemTests/Tests/WhenEnqueueActorMessagesViaHttpTests.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        await _enqueueActorMessagesHttpDsl
+            .EnqueueCalculatedMeasurementMessage(
+                _receiver,
+                meteringPointId);
```
ebbeknudsen, line 50: I know we only have one endpoint, but would it make sense to have 3 subsystem tests? (One for electrical heating, one for net consumption and one for capacity settlement. Atleast we should pick one, and rename the tests to use that name. Like `_enqueueActorMessages.EnqueueElectricalHeatingMessage()`. I'm not sure non techincal people know what "calculated measurements" are.

source/SubsystemTests/Tests/WhenEnqueueActorMessagesViaHttpTests.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        await _enqueueActorMessagesHttpDsl.ConfirmResponseIsAvailable(meteringPointId);
```
ebbeknudsen, line 52:
```suggestion
        await _enqueueActorMessages.ConfirmResponseIsAvailable(meteringPointId);
```

The purpose of the "DSL" is to remove technical terms from the code, so non technical personel can read and understand the tests, so lets keep the technical terms out of the test method.

PR #1614, opened 2025-05-12
test: Add subsystem test for enqueue missing measurements

ebbeknudsen:
## Description

Add subsystem test for enqueue missing measurements. Only tests that the endpoint works (for now), since the document writers hasn't been implemented yet.

## References

- https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/751
- D-002 run: https://github.com/Energinet-DataHub/dh3-environments/actions/runs/15019264533/job/42204863816

## Checklist
- [x] Reference to the task


source/SubsystemTests/Dsl/EnqueueActorMessagesHttpDsl.cs
```diff
@@ -110,4 +162,27 @@ private void AssertMeteringPointId(string content, string meteringPointId)
...
+            content);
+    }
+
+    private void AssertMeteringPointIdForRsm018(string content, string meteringPointId)
+    {
+        var meteringPointRegexPattern = MeteringPointRegexPattern.Replace(
+            "{__MeteringPointId__}",
+            meteringPointId);
+
+        Assert.Matches(new Regex(meteringPointRegexPattern), content);
```
RasmusGraabaek, line 186: Arn't this asserting the same strucuture as "AssertMeteringPointIdForRsm012"
If so, we might aswell keep it as a single method

ebbeknudsen: It is, but it is 2 different contracts/schemas, that just happens to write metering point id the same way, so the implementation could diverge. I decided to keep them separate, since they are 2 different schemas that just happens to look similar.

ebbeknudsen: And i introduced a regex instead of the method in `AssertMeteringPointIdForRsm012()`, since i do not like that it is reliant upon exactly the same whitespaces and newlines.

PR #1615, opened 2025-05-13
Test: Ensure archived messages are saved in expected table

RasmusGraabaek:
## Description
As of right now, we do not have a test which ensures that all the archived messages are stored in the expected table.
This pr introduces such a test

## Checklist
- [ ] Reference to the task

### DIFF
--- a/source/ArchivedMessages.IntegrationTests/WhenArchivedMessageIsCreatedTests.cs
+++ b/source/ArchivedMessages.IntegrationTests/WhenArchivedMessageIsCreatedTests.cs
@@ -94,6 +94,27 @@ public class WhenArchivedMessageIsCreatedTests : IAsyncLifetime
         Assert.NotNull(result);
     }
 
+    [Theory]
+    [MemberData(nameof(GetDocumentTypes))]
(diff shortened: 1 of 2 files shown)


source/ArchivedMessages.IntegrationTests/WhenArchivedMessageIsCreatedTests.cs
```diff
@@ -94,6 +94,27 @@ public async Task Given_ArchivedDocument_When_Created_Then_CanBeRetrieved(Docume
         Assert.NotNull(result);
     }
 
+    [Theory]
+    [MemberData(nameof(GetDocumentTypes))]
+    public async Task Given_DocumentType_When_Created_Then_StoredInExpectedDatabase(DocumentType documentType)
```
lasrinnil, line 99:
```suggestion
    public async Task Given_DocumentType_When_Created_Then_StoredInExpectedTable(DocumentType documentType)
```

lasrinnil (approved):

lasrinnil (approved):

PR #1665, opened 2025-06-11
feat: Add MeasurementSync endpoint (DO NOT MERGE)

lasrinnil:
## Description
- `MeasurementsSynchronizationTrigger` to handle receiving measurements from DH2, converting them to Ebix messages and forwarding them to the Process Manager.
- `MeasurementsJsonToEbixStreamWriter` now swaps sender and recipients, and removes the header from the written document.
- `MeasurementsJsonToEbixStreamWriterTests` now verifies the correct behavior with actually receiving the written document.

Should be merged as close to https://github.com/Energinet-DataHub/dh3-infrastructure/pull/3720/files as possible.
## Checklist

- [ ] Subsystem test executed [deploy to dev_002/dev_003](https://github.com/Energinet-DataHub/dh3-environments/actions/workflows/edi-cd.yml)


source/B2BApi.AppTests/Functions/EnqueueMessages/MeasurementsSynchronizationTriggerTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        var verifyServiceBusMessages = await _fixture.ServiceBusListenerMock
+            .When(msg =>
+            {
+                if (msg.Subject != "Brs_021_ForwardMeteredData")
+                    return false;
+
+                var parsedNotification = StartOrchestrationInstanceV1.Parser.ParseJson(
+                    msg.Body.ToString());
+
+                var matchingOrchestrationId = parsedNotification.MeteringPointId == "571051839308770693";
```
RasmusGraabaek, line 87:
```suggestion
                var matchingMeteringPointId = parsedNotification.MeteringPointId == "571051839308770693";
```

source/B2BApi/Functions/MeasurementsSynchronizationTrigger.cs
```diff
@@ -0,0 +1,69 @@
...
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
+using Microsoft.FeatureManagement;
+
+namespace Energinet.DataHub.EDI.B2BApi.Functions;
+
+/// <summary>
+/// Service Bus Trigger to process measurements synchronization from DataHub 2.
+/// </summary>
+public class MeasurementsSynchronizationTrigger(ILogger<MeasurementsSynchronizationTrigger> logger,
```
RasmusGraabaek, line 34:
```suggestion
public class MeasurementsSynchronizationTrigger(
ILogger<MeasurementsSynchronizationTrigger> logger,
```

source/IntegrationTests/Migration/MeasurementsJsonToEbixStreamWriterTests.cs
```diff
@@ -28,54 +33,23 @@ public class MeasurementsJsonToEbixStreamWriterTests
...
-            new MeasurementsToMarketActivityRecordTransformer());
+            serializer, new List<IDocumentWriter> { new MeteredDataForMeteringPointEbixDocumentWriter(new MessageRecordParser(serializer)) });
+        var meteredDataForMeteringPointEbixMessageParser = new MeteredDataForMeteringPointEbixMessageParser(
+            new EbixSchemaProvider(),
+            new Logger<MeteredDataForMeteringPointEbixMessageParser>(new LoggerFactory()));
 
         // Act
-        var stream = await timeSeriesJsonToEbixStreamWriter.WriteStreamAsync(JsonPayloadConstants.SingleTimeSeriesWithSingleObservation);
+        var timeSeriesPayload = Encoding.UTF8.GetBytes(JsonPayloadConstants.SingleTimeSeriesWithSingleObservation);
+        var stream = await timeSeriesJsonToEbixStreamWriter.WriteStreamAsync(new BinaryData(timeSeriesPayload));
```
RasmusGraabaek, line 45:
```suggestion
        var output = await timeSeriesJsonToEbixStreamWriter.WriteStreamAsync(new BinaryData(timeSeriesPayload));
```
`var stream` was confusing for me, since `item1` is the actual `stream`

source/B2BApi/MeasurementsSynchronization/MeasurementsJsonToEbixStreamWriter.cs
```diff
@@ -0,0 +1,95 @@
...
+        var stream = await _documentWriters.First(x => x.HandlesType(DocumentType.NotifyValidatedMeasureData) && x.HandlesFormat(DocumentFormat.Ebix)).WriteAsync(
+            outgoingMessageHeader,
+            meteredDataForMeteringPointMarketActivityRecords.Select(_serializer.Serialize).ToList(),
+            CancellationToken.None).ConfigureAwait(false);
+
+        // Load the stream into an XmlDocument for removing the header and formatting.
+        var xmlDoc = new XmlDocument();
+        xmlDoc.Load(stream.Stream);
+
+        var mgr = new XmlNamespaceManager(xmlDoc.NameTable);
```
RasmusGraabaek, line 68: Was `mgr` short of?

source/B2BApi/MeasurementsSynchronization/MeasurementsJsonToEbixStreamWriter.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        // Write the document using the MeteredDataForMeteringPointMarketActivityRecords and the outgoing message header
+        var stream = await _documentWriters.First(x => x.HandlesType(DocumentType.NotifyValidatedMeasureData) && x.HandlesFormat(DocumentFormat.Ebix)).WriteAsync(
+            outgoingMessageHeader,
+            meteredDataForMeteringPointMarketActivityRecords.Select(_serializer.Serialize).ToList(),
+            CancellationToken.None).ConfigureAwait(false);
+
+        // Load the stream into an XmlDocument for removing the header and formatting.
+        var xmlDoc = new XmlDocument();
+        xmlDoc.Load(stream.Stream);
```
ebbeknudsen, line 66: Missing disposing `stream`. When can it be disposed? After `xmlDoc.Load()`?

source/BuildingBlocks.Domain/FeatureManagement/FeatureFlagNames.cs
```diff
@@ -68,4 +68,9 @@ public static class FeatureFlagNames
     /// Whether to allow receiving CIM request messages for PM28.
     /// </summary>
     public const string PM28ReceiveCIMMessages = "PM28-CIM";
+
+    /// <summary>
+    /// Whether to allow synchronizing measurements from DataHub 2 to EDI.
+    /// </summary>
+    public const string SyncMeasurements = "SyncMeasurements";
```
ebbeknudsen, line 75: Shouldn't this be a release toggle instead, so we can activate/deactivate it in the Azure App Configuration UI?

source/B2BApi/Functions/MeasurementsSynchronizationTrigger.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            var streamAndSender = await _measurementsJsonToEbixStreamWriter.WriteStreamAsync(message.Body).ConfigureAwait(false);
+            _authenticatedActor.SetAuthenticatedActor(new ActorIdentity(ActorNumber.Create(streamAndSender.Sender), Restriction.Owned, ActorRole.GridAccessProvider, null,  null));
+            await _incomingMessageClient.ReceiveIncomingMarketMessageAsync(
+                    new IncomingMarketMessageStream(streamAndSender.Document),
+                    DocumentFormat.Ebix,
+                    IncomingDocumentType.NotifyValidatedMeasureData,
+                    DocumentFormat.Ebix,
+                    cancellationToken)
+                .ConfigureAwait(false);
```
ebbeknudsen, line 66: Just a thought (i also had when implementing Send Measurements B2C endpoint): If we updated our incoming messages client with a new method that received an `IIncomingMessage` (instead of a stream), then we wouldn't have to convert the entire data to a document, just to convert it back to an `IIncomingMessage`. This would also mean that we don't have to write custom parsers for our B2C requests, since we could just create the `IIncomingMessage` directly instead of writing a DTO to json, and then parsing the JSON into an `IIncomingMessage`.

The same goes for this instance, we wouldn't have to convert the JSON from migration to an ebIX document, we could instead just convert it to a new `MeteredDataForMeteringPointMessageBase : IIncomingMessage` (what our parsers return).

source/BuildingBlocks.Domain/FeatureManagement/FeatureFlagNames.cs
lasrinnil: Lets keep it as a setting.

source/B2BApi/MeasurementsSynchronization/MeasurementsJsonToEbixStreamWriter.cs
lasrinnil: "manager" short for XmlNameSpaceManager.

source/B2BApi/MeasurementsSynchronization/MeasurementsJsonToEbixStreamWriter.cs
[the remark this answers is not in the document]
lasrinnil: I have disposed the stream after loading it into the xml doc. This way we have the stream twice, and the deserialized data in scope.

ebbeknudsen (approved):

